PR #2, opened 2019-06-11
Fixes #1

### LINKED ISSUE #1: Authentication error when openstack-exporter fetches Cinder metrics
Hi,

At the original place where [this repository](https://github.com/Linaro/openstack-exporter/) was hosted someone created this [issue](https://github.com/Linaro/openstack-exporter/issues/9). I'm recreating this here instead since everything have moved.

I looked into the problem a bit and it seems that adding `volume` and not `volumev2` or `volumev3` breaks the whole logic in interesting ways. When the wrong service name, in this case the non-existing service `volume` (from an OpenStack perspective), is used when calling this method:

https://github.com/openstack-exporter/openstack-exporter/blob/e8f134c102410f53bf34c719161e3d07dce2d3f0/exporter.go#L95

And the subsequent authentication method call fires off here:

https://github.com/openstack-exporter/openstack-exporter/blob/e8f134c102410f53bf34c719161e3d07dce2d3f0/exporter.go#L96

Then we end up in this code block (from the forked upstream of `go-goose`):

https://github.com/niedbalski/goose/blob/2f08ad8bd5a896d24ee45b2cfc7c6ba2d3d744fc/client/client.go#L488-L497

If i trace it a bit further the `doAuthenticate()` done in the above code block have another method call which is this one:

https://github.com/niedbalski/goose/blob/2f08ad8bd5a896d24ee45b2cfc7c6ba2d3d744fc/client/client.go#L519-L521

If that call returns an error for some reason the following variables is not set on the `authenticatingClient` struct:
```
	c.apiURLVersions = make(map[string]*apiURLVersion)
	c.tenantId = authDetails.TenantId
	c.userId = authDetails.UserId
	// A valid token indicates authorisation has been successful, so it needs to be set last. It must be set
	// after the service URLs have been extracted.
	c.tokenId = authDetails.Token
```
Since the service name `volume` is "wrong" we get the following error message back (which is not propagated to the caller it seems):
```
INFO[0000] the configured region "RegionOne" does not allow access to all required services, namely: volume                                                   
access to these services is missing: volume
```
This is why i believe we get those `401`'s back from the Cinder API.

A quick fix would be to change `volume` to `volumev3`, and we're kind of assuming `v3` since we actually fetch and use the `v3` endpoint here:
https://github.com/openstack-exporter/openstack-exporter/blob/e8f134c102410f53bf34c719161e3d07dce2d3f0/cinder.go#L24

As i understand it the version 2 of the Block Storage API are deprecated, with that said there will probably be people using that older API still and there can be a need of supporting it.

Any feedback/input on this? Hopefully this was easy enough to follow.

I would love to contribute with e.g. a PR to start with, but i need some assistance, there's also things to fix in that upstream fork i guess.

mikejoh:
Fixes #1

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Flags:
       --disable-service.network  Disable the network service exporter
       --disable-service.compute  Disable the compute service exporter
       --disable-service.image    Disable the image service exporter
-      --disable-service.volume   Disable the volume service exporter
+      --disable-service.volumev3   Disable the volumev3 service exporter
       --disable-service.identity
                                  Disable the identity service exporter
 
--- a/exporter.go
+++ b/exporter.go
@@ -117,7 +117,7 @@ func NewExporter(name string, prefix string, config *Cloud) (OpenStackExporter, 
 				return nil, err
 			}
 		}
-	case "volume":
+	case "volumev3":
 		{
 			exporter, err = NewCinderExporter(newClient, prefix, config)
 			if err != nil {
--- a/main.go
+++ b/main.go
@@ -24,7 +24,7 @@ func EnableExporter(service string, prefix string, config *Cloud) (*OpenStackExp
 	return &exporter, nil
 }
 
-var defaultEnabledServices = []string{"network", "compute", "image", "volume", "identity"}
+var defaultEnabledServices = []string{"network", "compute", "image", "volumev3", "identity"}
 
 func main() {
 	var (


mikejoh:
I just realized that there are tests missing for Cinder in `exporter_test.go` and of course fixtures. That could also be something to work out, me doing this the wrong way perhaps.

mikejoh:
I may have messed something up when rewording the commit message, sorry for any confusion (if any)!

mikejoh:
After thinking about this PR and the issue a bit i’ll redo it and send in another one!

niedbalski:
OK, i will wait for your fix @mikejoh , what's the issue you are trying to overcome?

PR #3, opened 2019-06-13
Fix cinder authentication and error handling when enabling exporters

mikejoh:
For [#1]

This PR addresses the issue with authenticating to the cinder/block storage API which breaks due to authentication against a non-existing service `volume`.

I've also added some error handling when authentication happens before creating specific exporter, now we get the error(s) propagated back to the caller. 

If there's an error enabling a specific exporter `openstack-exporter` wont exit with a fatal error but instead log the error and continue with the next exporter.

Worth mentioning is that part of this PR can be broken into other PRs against the upstream repository with the main logic of talking to the different OpenStack APIs.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -93,7 +93,9 @@ func NewExporter(name string, prefix string, config *Cloud) (OpenStackExporter, 
 	}
 
 	newClient.SetRequiredServiceTypes([]string{name})
-	newClient.Authenticate()
+	if err := newClient.Authenticate(); err != nil {
+		return nil, fmt.Errorf("error when authenticating: %s", err)
+	}
 
 	switch name {
 	case "network":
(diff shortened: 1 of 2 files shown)


niedbalski:
Hello @mikejoh 

Many thanks for your contributions and for using this project. I have merged your changes
into the master branch, however, I won't be releasing a new version until fixing some remaining
issues with the ci (#6 ) that I hope to have fixed very soon.

Again, thanks for contributing and taking the time to fix this :-)

Cheers!

niedbalski:
Hello @mikejoh 

This fix has been released as part of v0.0.8 and you have been added to the CONTRIBUTORS section.

mikejoh:
> This fix has been released as part of v0.0.8 and you have been added to the CONTRIBUTORS section.

Thanks @niedbalski, that sounds great, glad to be on-board! :-)

PR #9, opened 2019-06-19
config: make sure defaults don't overwrite set values

pborzenkov:
Default config values must be applied *before* unmarshalling the actual
config, otherwise they will always overwrite set values.


niedbalski:
LGTM, thanks for your contribution @pborzenkov added to 0.0.9 milestone.

PR #14, opened 2019-07-15
Changed method for cinder client to getting current information about…

1pintbeer:
Changed method for cinder client to getting current information about volumes size.
GetVolumesSimple does't return info about volume size. GetVolumesDetail return all volume details.

### DIFF
--- a/cinder.go
+++ b/cinder.go
@@ -87,7 +87,7 @@ func (exporter *CinderExporter) Describe(ch chan<- *prometheus.Desc) {
 
 func (exporter *CinderExporter) Collect(ch chan<- prometheus.Metric) {
 	log.Infoln("Fetching volumes info")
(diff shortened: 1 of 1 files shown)


1pintbeer:
Yes, GetVolumesSimple does't return volume size. GetVolumesSimple lists summary information for all Block Storage. GetVolumesDetail lists detailed information for all Block Storage. When we collect metrics, we get -1 value for openstack_cinder_volume_status metric.

1pintbeer:
> Are you seeing any error or missing data? Where are you using this function call return?

Yes, GetVolumesSimple does't return volume size. GetVolumesSimple lists summary information for all Block Storage. GetVolumesDetail lists detailed information for all Block Storage. When we collect metrics, we get -1 value for openstack_cinder_volume_status metric.

niedbalski:
Thank you @1pintbeer ,  LGTM.

PR #27, opened 2019-09-24
Move from goose to gophercloud

alistarle:
Here is the draft of my pull request to move from goose to gophercloud, a lot of work remain to do :

- [ ] Use the --os-client-config args (currently, loading of the clouds.yml are done trough this [function](https://github.com/gophercloud/utils/blob/34f5991525d116b3832e0d9409492274f1c06bda/openstack/clientconfig/utils.go#L99))
- [ ] Check that all the client really use the params from the clouds.yml (endpoint type, tls, etc...)
- [ ] Fix the tenant_id of the volume_status metrics (extended attribute of the volume API)
- [ ] Check that the refresh_client really works (make tests)
- [ ] Compare scrape_duration between the goose version and the gophercloud one
- [ ] Factorize some code in the exporter
- [ ] Write/Fix the unit tests


niedbalski:
Hello @alistarle 

I have created a feature base branch to work on this implementation, the branch can be found here [0] under the name "gophercloud"

It has already covered some of the points you mentioned in the TODO list, including; use the os-client-config option and check that all the client really use the params from the clouds.yml 

* Could you please rebase your changes on top of that branch and submit the PR against it so we can collaborate from there rather than PR on top of master?

Also, please ignore any changes into the vendors dependencies, we can add those dependencies later via go mod.

[0] https://github.com/openstack-exporter/openstack-exporter/tree/gophercloud

PR #35, opened 2019-10-07
Add availability zone information to nova metrics

alistarle:
Now we are using Gophercloud, we are able to add more complex metrics, I open the dance with the missing "aggregate" information on nova metrics !


alistarle:
Note that we need to know what to do if a host is in multiple aggregate : Should we return the first one ? The last one ? Create one metrics per aggregate ?
If no solution are acceptable, maybe we need to return only the availability zone label, as it is more unique per host I think

exporters/nova.go
```diff
@@ -129,30 +131,48 @@ func ListHypervisors(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metri
...
+		return
+	}
+
+	hostToAggregateMap := map[string]string{}
+	for _, v := range allAggregates {
+		for _, w := range v.Hosts {
+			hostToAggregateMap[w] = v.AvailabilityZone
+		}
+	}
+
```
niedbalski: I think it would be better to generate a metric per aggregate or Join the list of aggregates by comma

exporters/nova.go
```diff
@@ -129,30 +131,48 @@ func ListHypervisors(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metri
...
 	for _, hypervisor := range allHypervisors {
 		ch <- prometheus.MustNewConstMetric(exporter.Metrics["running_vms"].Metric,
-			prometheus.GaugeValue, float64(hypervisor.RunningVMs), hypervisor.HypervisorHostname, "")
+			prometheus.GaugeValue, float64(hypervisor.RunningVMs), hypervisor.HypervisorHostname, hostToAggregateMap[hypervisor.HypervisorHostname])
```
niedbalski: aggregate := ""
if val, ok := hostToAggregate[hypervisor.HypervisorHostname]; ok {
    aggregate = val
}

niedbalski:
@alistarle 

Thank you, can you check the inline comments in left you in the PR?

alistarle:
I thinked a lot about the two option, and I don't yet know the best one as:
- Return multiple metrics can be a bad idea, as by example if a host is in multiple aggregate, and you query `sum(vcpus_available)` you will count the same host multiple time
- Return one metric separated by coma is maybe better, but I don't know if you can handle it easily in PromQL ?

That's why I propose to move this PR to return the availability zone instead of the aggregate, which is easier as one host can be part only of one availablity zone at a time. Maybe we can look at the aggregate metric after ?

niedbalski:
LGTM

PR #49, opened 2019-10-25
Add 6 missing server_status to nova exporter"

tzmtl:
I added 6 missing server_status to nova exporter. 
https://docs.openstack.org/api-guide/compute/server_concepts.html

I didn't delete any thing. I didn't change the order of old server_status too, to make sure this commit will not break the configurations of other users. 

It's my editor that aligned all comments :)


niedbalski:
Thank you @tzmtl !

PR #53, opened 2019-10-30
Make network_ip_availabilities metric per subnet

alistarle:
In order to complete the openstack_neutron_network_ip_availabilities_used/total metrics, I propose to render it per subnet, as we already have the information, and it cover the use case where you have multiple subnet per networks.


niedbalski:
LGTM

Issue #55, opened 2019-10-30
Add ability to disable metric gathering


alistarle:
In order to limit impact of the exporter againt large Openstack environment, it will be very usefull to be able to blacklist some metrics, by example the "openstack_nova_server_status" or the "openstack_cinder_volume_status".

We can take example on the node-exporter who use "--no-collector.<name>" argument, or using a configuration file.

niedbalski:
@alistarle we have the '--disable-*exporter*' options, are you thinking something more granular per-metric?

alistarle:
Yes, maybe a --disable-metric option so, because using this exporter in an deployment with Manu thousands of VM or volumes is impossible for now I think

niedbalski:
@alistarle happy to review or work on this feature.

niedbalski:
Hello @alistarle 

I have implemented this as part of #56, please let me know if that change is acceptable for you before proceeding to close this ticket.

Thanks!

PR #56, opened 2019-11-08
Add --disable-metric parameter.

niedbalski:
This commit adds a new parameter --disable-metric (short d)
that allows to specify a metric to be skip from collection.

The format has to be --disable-metric service_name-metric,
some examples are:

--disable-metric cinder-snapshots
--disable-metric nova-servers

This issue refers to #55


alistarle:
Seems ok for me, thanks for implementing it !

PR #69, opened 2019-12-06
Add $service_up values

mnaser:
This adds service_up values in order for us to be able to check if the API is responding successfully, this can be helpful in knowing that the endpoints are not healthy.

### DIFF
--- a/exporters/exporter_test.go
+++ b/exporters/exporter_test.go
@@ -125,6 +125,13 @@ func (suite *NovaTestSuite) TestNovaExporter() {
 	for _, metric := range defaultNovaMetrics {
 		suite.Contains(suite.Recorder.Body.String(), "nova_"+metric.Name)
 	}
+
+	suite.Contains(suite.Recorder.Body.String(), "nova_up 1")
+}
+
+func (suite *NovaTestSuite) TestNovaExporterWithEndpointDown() {
+	suite.StartMetricsHandler()
+	suite.Contains(suite.Recorder.Body.String(), "nova_up 0")
 }
 
 type NeutronTestSuite struct {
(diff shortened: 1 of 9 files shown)


mnaser:
Ended up doing a few more clean-ups too but this should report happily if a service is up or not :)

exporters/exporter_test.go
```diff
@@ -126,9 +136,25 @@ func (suite *NovaTestSuite) TestNovaExporter() {
 
 	suite.StartMetricsHandler()
 
+	for _, fam := range suite.MetricFamilies {
+		fmt.Println(fam)
+	}
+	// fmt.Println(suite.MetricFamilies)
```
niedbalski: Not needed.

exporters/exporter_test.go
```diff
@@ -126,9 +136,25 @@ func (suite *NovaTestSuite) TestNovaExporter() {
...
+
 	for _, metric := range defaultNovaMetrics {
-		suite.Contains(suite.Recorder.Body.String(), "nova_"+metric.Name)
+		fmt.Println(suite.MetricFamilies["openstack_nova_"+metric.Name])
```
niedbalski: Not needed.

exporters/keystone.go
```diff
@@ -40,115 +39,94 @@ func NewKeystoneExporter(client *gophercloud.ServiceClient, prefix string, disab
...
 
-func (exporter *KeystoneExporter) Describe(ch chan<- *prometheus.Desc) {
-	for _, metric := range exporter.Metrics {
-		ch <- metric.Metric
-	}
-}
-
-func ListDomains(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metric) {
-
-	log.Infoln("Fetching domains information")
```
niedbalski: No need to remove log lines.

exporters/exporter_test.go
mnaser: Woops, debugging stuff, gone.

exporters/exporter_test.go
mnaser: I was using it  for debugging, gone!

exporters/keystone.go
mnaser: So the reason I removed it was because we pretty much have the same behavior here:

https://github.com/openstack-exporter/openstack-exporter/blob/master/exporters/exporter.go#L79

It's logging the same exact info pretty much?

niedbalski:
LGTM

PR #73, opened 2019-12-16
Add Nova limits metrics

prolane:
### This PR aims to add metrics to expose nova limits.

**Why?**
We would like to monitor if tenants run out of their quotas/limits. Also, on the long run, we would like to use this for capacity planning. Meaning: Compare the total assigned quota with the total available capacity on the hypervisors. The latter is already available as metric, the first one however is not.

**Remarks**
I'm using [compute/limits](https://godoc.org/github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/limits) to expose these metrics, which as you can see is part of the nova API. Therefore, I've added these metrics to the nova exporter. The one needed variable for the `limits.Get` function however is the `TenantID`. Therefore, I had to query the list of projects using the Identity/Keystone gophercloud.ServiceClient. As the `NovaExporter` only holds a nova client, I had to think of a way to obtain a identity client. I achieved this by simply storing all clients in a map.  I had several scenarios in mind to solve this 'issue', but in the end I think this is the most elegant one, without having to rewrite large pieces of code. I hope you can appreciate this solution.


prolane:
I see the CircleCI test failed. I'll look at the CircleCI output later today.

prolane:
The test fails, because if understand the test suite correctly, when `TestNovaExporter()` runs, only the nova exporter is started.

As described in the PR description, we need to get a list of all projects/tenants to be able to query the nova limits for each project. I'm therefore depending on the identity/keystone exporter to be up, as I'm re-using the gophercloud client for the identity API. See these lines of code:
* [Store all created clients in a map](https://github.com/springernature/openstack-exporter/blob/68bac46abf3dfb5ae245017f3c05b8f27051841a/exporters/exporter.go#L165-L168)
* [Use the identity client](https://github.com/springernature/openstack-exporter/blob/68bac46abf3dfb5ae245017f3c05b8f27051841a/exporters/nova.go#L275-L277)

This works fine, but as said, the go test fails on this as the identity exporter is not available.

I can work around this, by just creating a new client like this:

```
c, err := openstack.NewIdentityV3(exporter.Client.ProviderClient, gophercloud.EndpointOpts{Region: "RegionOne"})
if err != nil {
    return err
}

allPagesProject, err := projects.List(c, projects.ListOpts{}).AllPages()
```

This will work fine for most users and the go tests are all successful as well. However, this should be improved as the `Region` is hardcoded and the `Availability` is not set. I could rewrite some stuff so that the `gophercloud.EndpointOpts` can be obtained by calling a function from `utils` for example? What are your thoughts?

prolane:
With the code I just pushed, the issue is solved and the go tests are successful as well. 

I think things could be more elegant, but this would require some rewriting of code. My idea would be to read the config (e.g. the clouds.yml) just once at startup and store this data in a global variable. Then it would be possible to call the `NewServiceClient` function from anywhere.

Anyway, perhaps for a next iteration.

niedbalski:
LGTM, thank you @prolane for your contribution. Great addition!

Issue #74, opened 2019-12-19
Add 'aggregate' label to nova metrics


prolane:
Hi!
I'd like to open up a former discussion about adding the 'aggregate' label to some nova metrics. @alistarle initially created a PR for this over here: https://github.com/openstack-exporter/openstack-exporter/pull/35. Because of the discussions about how to best implement this, the PR ended in adding an AZ label instead (still very useful btw, thank you for adding).

I would like to start using the `aggregate` label though, therefore I'd like to discuss this again. I've been thinking about what the best approach would be and here is my proposal:

### Proposal
How about we make the `aggregate` label optional by setting a flag when starting the exporter? If you explicitly set this aggregate flag, the way the nova metrics are generated changes a bit. Instead of looping over all hypervisors, we would be first looping over all aggregates and then for each aggregate loop over all hypervisors in that aggregate.

I do understand like @alistarle commented in https://github.com/openstack-exporter/openstack-exporter/pull/35, you would lose the possibility to query `sum(vcpus_available)` (as one hypervisor is potentially part of more than one aggregate). But this is why I think enabling the aggregate information **optionally** with a flag makes sense, as you then explicitly accept the changed behavior. I don't think people who work with aggregates want to see the sum of `vcpus_available` over all hypervisors. What they do want to see is the sum of `vcpus_available` grouped by aggregate. Which will be possible with the `aggregate` label implemented as suggested.

Let me know what you guys think. I'll be happy to create a PR for it.

niedbalski:
@prolane 

In my opinion a single label named 'aggregates', which can be a space/comma separated list of aggregates on which a specific hypervisor belongs to should suffice to solve this issue as you can filter by a specific label and sum the vcpus_available. Any drawbacks on using this approach?

prolane:
Thats certainly a possibility @niedbalski. I guess if we make the space/comma separated list in alphabetical order, we make sure we don't end up with lots of different combinations. For example, you don't get one hypervisor which has the 'aggregates' label value set to `"az-west,ssd-hosts,shared-cluster"` and one other set to `"ssd-hosts,az-west,shared-cluster"`.

Just one drawback of this approach is, if you have hypervisors being part of more than one aggregate, you can not group by one aggregate. For example, with the example aggregates label values above, I won't be able to group by just the `ssd-hosts`. Not without making hard coded dashboards in Grafana for example.

Ideally, you would want to do something like this:
`sum(vcpus_available) by(aggregates)`

[...]

PR #75, opened 2019-12-23
Add 'aggregates' label to nova metrics

prolane:
This PR adds the `aggregates` label to the nova metrics. As hosts/hypervisors can be part of more than one aggregate, the `aggregates` label is a comma separated list of the aggregates to which the host belongs to. (As discussed in https://github.com/openstack-exporter/openstack-exporter/issues/74)

It ignores aggregates which are created with the sole purpose of placing a host in an availability zone, as this information is already captured by the `availability_zone` label.


niedbalski:
LGTM

PR #77, opened 2020-01-11
Improve testing and add ip_version to availabilities

mnaser:
This change improves the testing for the exporter, now testing actual values!  It also fixes a few missing mocks and makes the test fail if there is any missing mocks.

It also adds ip_version which is really useful for checking for IP availability.

### DIFF
--- a/exporters/cinder_test.go
+++ b/exporters/cinder_test.go
@@ -1,0 +1,52 @@
+package exporters
+
+import (
+	"strings"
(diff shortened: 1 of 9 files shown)


niedbalski:
The code looks good to me, I think is genuinely an awesome improvement to the test suite for making testing a lot easier.

Thank you @mnaser for this amazing contribution.

PR #82, opened 2020-02-07
New Neutron metrics + inconsistency checks

Hybrid512:
- Added basic loadbalancer metrics
- Added metrics for inconsistencies checks (ports without IPs, ...)

These metrics were needed for our monitoring of our clusters (we have about a dozen clusters for all kind of environments) but they might be useful to a broader audience.


niedbalski:
LGTM @Hybrid512

PR #93, opened 2020-03-26
Fix a bug can cause duplicate metrics and improve the implementation

tzmtl:
In the original implementation when there are more than 10000 objects in a container,
there will be mutiple pages of objects list, the label container_name is the same
one, so it will cause an error of duplicate metrics.

And we don not have to list all objects in a container and count to know how many
objects are in a container. We can get this from container attribute


tzmtl:
The current implementation still needs improvements. Now it can only get metrics of containers of the account that authenticates with Keystone. It can not list containers of all accounts. No much sense. But I didn't find the method to list containers of all accounts in Go Doc. We will see later how to do implement this.

niedbalski:
@tzmtl thank you for your contribution, could you please review the failing tests and add the "bytes" property in the fixtures?

PR #97, opened 2020-04-12
Performance improvements + port status addition

mnaser:
The neutron exporter seems to have a lot of "active/inactive" references
which are doing a query every single time, resulting in duplicate API
calls on very heavy operations (aka list all routers/ports in a cloud).

This PR updates testing to allow calling a URL only *once* during
tests which should eliminate these from making their way into the
code.

It also introduces a listing of all ports so a user can opt into
checking things like their binding state, which can be key in finding
issues in Neutron ports being unplugged.

This potentially generates a lot of records, but users can use
recording rules to aggregate them.


niedbalski:
LGTM, thank you once again @mnaser for this contribution.

Issue #106, opened 2020-07-05
Metrics of DNS zones only for current project


alexeymyltsev:
For this moment zones and recordsets collecting only for current project.
For collection information about zones in all projects needed add to headers "x-auth-all-projects"(https://docs.openstack.org/api-ref/dns/?expanded=list-zones-detail,list-recordsets-in-a-zone-detail). 
It is possible if add code like this
```client.MoreHeaders = map[string]string{"X-Auth-All-Projects": "True"}```
but as far as I understand this header will be used for requests to another services too. 

May be we can do deepcopy of client just for Designate metrics?
Could you advice how would be better solve this problem?

PR #108, opened 2020-07-06
Fix dns metrics for all projects

alexeymyltsev:
This PR add header x-auth-all-projects: true for client. This header allow collect information about zones from all projects.
It fix issue: https://github.com/openstack-exporter/openstack-exporter/issues/106


alexeymyltsev:
I have checked dumps of request and with this PR headers using with all request. With PR https://github.com/openstack-exporter/openstack-exporter/pull/109 there is no additional headers with non designate requests.

PR #113, opened 2020-07-14
Add Ironic node state exporter

diurnalist:
This adds a new exporter that scrapes the state of Ironic nodes in the
deployment and returns their UUIDs, names, provision and power states,
maintenance states, and console enabled states.

This also has to add the baremetal service exporter by default, as there
is no way to explicitly enable a service that defaults to disabled. To
disable, use the --disable-service.baremetal flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# OpenStack Exporter for Prometheus [![Build Status][buildstatus]][circleci] 
+# OpenStack Exporter for Prometheus [![Build Status][buildstatus]][circleci]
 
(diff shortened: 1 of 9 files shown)


diurnalist:
Sorry about the whitespace changes, I left them in because they seemed reasonable and were auto-generated by my editor. Happy to remove if you want--I was lazy and wanted to check if you're OK with them first :)

niedbalski:
@diurnalist Hello Jason. 

 LGTM. Thank you so much for taking the time to propose this patch :-). I'll merge it.

Issue #119, opened 2020-08-13
Host name is missing in openstack_nova_server _status 


Jerin-Altnix:
Team,

We are unable to find the hostname of the compute node in the openstack_nova_server _status 
Kindly guide us, so that we can find in which compute node the VM is been deployed

Sample: 
openstack_nova_server_status{address_ipv4="",address_ipv6="",availability_zone="nova",flavor_id="2",host_id="9195214f95b39b8ecf27de26e32424ec6339a18cb9e0831ae8e3668c",id="65b0542e-4df6-4b54-a5eb-d29a7d0194f0",name="centos-test3",status="SHUTOFF",tenant_id="f450607bdf494f3da265ba61934638d6",user_id="30bb28e6cb394b10bbf8fc9a37c40aa7",uuid="65b0542e-4df6-4b54-a5eb-d29a7d0194f0"} 11

As per the official document in GitLab, there is a column called sample labels and it contains hostname in example 

Thanks in advance

PR #123, opened 2020-08-24
Add hypervisor_hostname label to Nova server_status metric

mikejoh:
By accessing the extended server attributes (added in [Kilo](https://docs.openstack.org/nova/queens/reference/api-microversion-history.html#maximum-in-kilo)) we can add the `hypervisor_hostname` label to the Nova `server_status` metrics.

Does this make sense? I thought it would be a nice addition, the `host_id` label which is the hypervisor host id was the main reason for me looking to add something else (which didn't require extra lookups) to query for.


mikejoh:
Fixes #119

niedbalski:
@mikejoh LGTM

PR #124, opened 2020-09-18
Metrics for each Floating IP

alexeymyltsev:
For monitoring NAT rules on network nodes we would like to collect information about each floating IP and route on which it located.


niedbalski:
@alexeymyltsev as always, great contribution. Thank you!.

Issue #127, opened 2020-09-25
Loadbalancer metric should be part of loadbalancer.go


johscheuer:
The `lodabalancer` metric provided in https://github.com/openstack-exporter/openstack-exporter/blob/master/exporters/neutron.go#L40-L41 should actually be part of the https://github.com/openstack-exporter/openstack-exporter/blob/master/exporters/loadbalancer.go exporter otherwise you get the following error when you run the openstack exporter with `openstack-exporter --disable-service.load-balancer --disable-service.container-infra` in an environment without LBaaS you get the following error:

```bash
... metric for exporter: neutron, error: failed to collect metric: loadbalancers, error: Resource not found" source="exporter.go:128"
```

we ended up by adding the `--disable-metric=loadbalancers`. I think for a better usage the metric should be moved.

PR #128, opened 2020-09-25
Remove loadbalancer metrics from neutron

johscheuer:
Fixes: https://github.com/openstack-exporter/openstack-exporter/issues/127

I just saw that there are some duplications, in the end we cloud drop both metrics defined in `neutron`:

- `loadbalancers` is the same as `total_loadbalancers`
- `loadbalancers_not_active` can be calculated from `loadbalancer_status`

I don't know if there are some deprecation policies but from my point of view we can drop both metrics.


niedbalski:
LGTM, thank you so much for this contribution @johscheuer

PR #131, opened 2020-10-07
Add openstack_identity_project_info metric

jocelynthode:
This PR adds the openstack_identity_project_info to be able to retrieve various info like tenant_id/description from a metric.

It also adds testing for the keystone component

### DIFF
--- a/exporters/keystone.go
+++ b/exporters/keystone.go
@@ -1,6 +1,8 @@
 package exporters
 
 import (
+	"strconv"
+
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/domains"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/groups"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
@@ -18,6 +20,7 @@ var defaultKeystoneMetrics = []Metric{
 	{Name: "users", Fn: ListUsers},
 	{Name: "groups", Fn: ListGroups},
 	{Name: "projects", Fn: ListProjects},
+	{Name: "project_info", Labels: []string{"is_domain", "description", "domain_id", "enabled", "id", "name"}},
 	{Name: "regions", Fn: ListRegions},
 }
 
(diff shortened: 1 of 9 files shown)


jocelynthode:
@niedbalski Hi, let me know if you need anything else in this  PR

exporters/keystone.go
```diff
@@ -69,6 +72,11 @@ func ListProjects(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metric)
 
 	ch <- prometheus.MustNewConstMetric(exporter.Metrics["projects"].Metric,
 		prometheus.GaugeValue, float64(len(allProjects)))
+	for _, p := range allProjects {
+		ch <- prometheus.MustNewConstMetric(exporter.Metrics["project_info"].Metric,
+			prometheus.GaugeValue, 1.0, strconv.FormatBool(p.IsDomain),
```
niedbalski, line 77: Why not setting the metric value to p.Enabled or not (1, or 0) rather than a static 0?

niedbalski:
Hello @jocelynthode 

Thank you so much for this contribution, for taking care of appending the README.md, adding tests
and fixtures. I just had a single comment that I pointed inline.  Once we figure out that discrepancy, we
I am good to merge this request.

exporters/keystone.go
jocelynthode: I exposed a `1` here as I consider this metric a "pseudo-metric" hence the suffix `_info`. I chose this value based on the prometheus best practices documentation (https://prometheus.io/docs/practices/naming/#metric-names), which points to this article suggesting a value of 1 for pseudo-metrics (https://www.robustperception.io/exposing-the-software-version-to-prometheus).

This is rather useful because you can then use this pseudo metric to add labels to a given query by taking advantage of the `*` operator.

Would this be ok with you?

jocelynthode:
@niedbalski Thanks for the quick review! Yes I noticed that the project had no tests and fixtures for keystone while adding the new metric so I had a bit of fun adding them in :)

exporters/keystone.go
jocelynthode: Thinking about it, it might be a good idea to have an other metric like `openstack_identity_project_enabled` which we could set to either `0` or `1` and keep this info out of the `project_info` metric? As I'm rather new to prometheus let me know if this is a terrible idea :)

niedbalski: OK, I think I agree with the idea of leaving this one as a pseudo-metric.

PR #140, opened 2020-11-07
Add support for proxy

### LINKED ISSUE #117: https proxy support
Hello,
im trying to use the exporter from our corporate network, with no direct access to Openstack API in internet.
I tried setting https_proxy env var but i guess this needs an evolution of the exporter for proper proxy support ?
Thks for your feedback

rustycl0ck:
closes #117

### DIFF
--- a/exporters/exporter.go
+++ b/exporters/exporter.go
@@ -198,10 +198,6 @@ func NewExporter(name, prefix, cloud string, disabledMetrics []string, endpointT
 		transport = &http.Transport{TLSClientConfig: tlsConfig}
 	}
 
-	if transport != nil {
-		transport.Proxy = http.ProxyFromEnvironment
-	}
-
 	client, err := NewServiceClient(name, &opts, transport, endpointType)
 	if err != nil {
 		return nil, err
--- a/exporters/exporter.go
+++ b/exporters/exporter.go
@@ -3,12 +3,12 @@ package exporters
 import (
 	"crypto/tls"
 	"fmt"
-	"github.com/hashicorp/go-uuid"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


exporters/exporter.go
```diff
@@ -198,6 +198,10 @@ func NewExporter(name, prefix, cloud string, disabledMetrics []string, endpointT
 		transport = &http.Transport{TLSClientConfig: tlsConfig}
 	}
 
+	if transport != nil {
```
niedbalski, line 201: is this validation required? could be new()'ed before?

niedbalski:
LGTM, quick observation.

niedbalski:
@rustycl0ck any feedback on this?

exporters/exporter.go
rustycl0ck: yes earlier I tried to initialize it above the bock:
```diff
+	transport.Proxy = &http.Transport{http.ProxyFromEnvironment}
	if !*config.Verify {
		log.Infoln("SSL verification disabled on transport")
		tlsConfig := &tls.Config{InsecureSkipVerify: true}
-		transport = &http.Transport{TLSClientConfig: tlsConfig}
+		transport.TLSClientConfig = tlsConfig
	}
```

but didn't know if it would have any consequences for cases where `config.Verify` is `true` or for non SSL cases, because the `transport` would then always be non-nil in the below line 
```
client, err := NewServiceClient(name, &opts, transport, endpointType)
```

rustycl0ck:
@niedbalski sorry for the late response. I don't think I'll be able to spend time on testing/modifying this in the immediate future. I see that @ichtar has been trying it out. So I'll hand it over to them for now until I can spend time on this again.

niedbalski:
LGTM

PR #141, opened 2020-11-08
Add support for multiple clouds

### LINKED ISSUE #139: [Feature Request] Support for multiple clouds?
Right now there is an option to choose only a single cloud from the `clouds.yaml` file. Is there any plan to make it possible to scrape multiple openstack instances through the same exporter?

I know running multiple exporters might be a solution, but in my scenario, I need to scrape around 50 or more openstack instances, thus running so many exporters isn't really easily manageable.

rustycl0ck:
closes #139 

:warning: PR in `Draft` status currently because it breaks the `--collect-metric-time` functionality with error:
```
* collected metric openstack_metric_collect_seconds label:<name:"openstack_metric" value:"regions" > label:<name:"openstack_service" value:"openstack_identity" > gauge:<value:1.845333353 >  with unregistered descriptor Desc{fqName: "openstack_metric_collect_seconds", help: "Time needed to collect metric from OpenStack API", constLabels: {openstack_metric="regions",openstack_service="openstack_identity"}, variableLabels: []}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,20 +53,22 @@ and must by specified with the `--os-client-config` flag.
 
 Other options as the binding address/port can by explored with the --help flag.
 
-By default the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+By default the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL.
 
 You can build it by yourself by cloning this repository and run:
 
 ```sh
 make common-build
-./openstack-exporter --os-client-config /etc/openstack/clouds.yaml region.mycludprovider.org
+./openstack-exporter --os-client-config /etc/openstack/clouds.yaml
+curl "http://localhost:9180/probe?cloud=region.mycludprovider.org"
 ```
 
 Or alternatively you can use the docker images, as follows (check the openstack configuration section for configuration
 details):
 
 ```sh
-docker run -v "$HOME/.config/openstack/clouds.yml":/etc/openstack/clouds.yaml -it quay.io/niedbalski/openstack-exporter-linux-amd64:master my-cloud.org
+docker run -v "$HOME/.config/openstack/clouds.yml":/etc/openstack/clouds.yaml -it -p 9180:9180 quay.io/niedbalski/openstack-exporter-linux-amd64:master
+curl "http://localhost:9180/probe?cloud=my-cloud.org"
 ```
 
 ### Command line options
@@ -88,27 +90,30 @@ Flags:
       --endpoint-type="public"   openstack endpoint type to use (i.e: public, internal, admin)
       --collect-metric-time      time spent collecting each metric
   -d, --disable-metric= ...      multiple --disable-metric can be specified in the format: service-metric (i.e: cinder-snapshots)
-      --disable-service.network  Disable the network service exporter
-      --disable-service.compute  Disable the compute service exporter
-      --disable-service.image    Disable the image service exporter
-      --disable-service.volume   Disable the volume service exporter
-      --disable-service.identity
-                                 Disable the identity service exporter
-      --disable-service.object-store
-                                 Disable the object-store service exporter
-      --disable-service.load-balancer
-                                 Disable the load-balancer service exporter
-      --disable-service.container-infra
-                                 Disable the container-infra service exporter
-      --disable-service.dns      Disable the dns service exporter
-      --disable-service.baremetal
-                                 Disable the baremetal service exporter
-      --disable-service.gnocchi  Disable the gnocchi service 
       --disable-slow-metrics     disable slow metrics for performance reasons
       --version                  Show application version.
+```
 
-Args:
-  <cloud>  name or id of the cloud to gather metrics from
+### Scrape options
+
+Which cloud (name or id from the `clouds.yaml` file) or what services from the cloud to scrape, can be specified as the parameters to http scrape request.
+
+Query Parameter | Description
+--- | ---
+`cloud` | Name or id of the cloud to gather metrics from (as specified in the `clouds.yaml`)
+`include_services` | A comma separated list of services for which metrics will be scraped. Defaults to all services: "network,compute,image,volume,identity,object-store,load-balancer,container-infra,dns,baremetal,gnocchi"
+`exclude_services` | A comma separated list of services for which metrics will *not* be scraped. Default is empty: ""
+
+Examples:
+```
+## Scrape all services from `test.cloud`
+curl "https://localhost:9180/probe?cloud=test.cloud"
+
+## Scrape only `network` and `compute` services from `test.cloud`
+curl "https://localhost:9180/probe?cloud=test.cloud&include_services=network,compute"
+
+## Scrape all services except `load-balancer` and `dns` from `test.cloud`
+curl "https://localhost:9180/probe?cloud=test.cloud&exclude_services=load-balancer,dns"
 ```
 
 ### OpenStack configuration
--- a/README.md
+++ b/README.md
@@ -53,7 +53,9 @@ and must by specified with the `--os-client-config` flag.
 
 Other options as the binding address/port can by explored with the --help flag.
 
-By default the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL.
+The exporter can operate in 2 modes
+- A Legacy mode (targetting one cloud) in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+- A multi cloud mode in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL.
 
 You can build it by yourself by cloning this repository and run:
 
@@ -89,13 +91,36 @@ Flags:
       --prefix="openstack"       Prefix for metrics
       --endpoint-type="public"   openstack endpoint type to use (i.e: public, internal, admin)
       --collect-metric-time      time spent collecting each metric
-  -d, --disable-metric= ...      multiple --disable-metric can be specified in the format: service-metric (i.e: cinder-snapshots)
       --disable-slow-metrics     disable slow metrics for performance reasons
+      --all-clouds               Toggle the multiple cloud scrapping mode under /probe?cloud=
+  -d, --disable-metric= ...      multiple --disable-metric can be specified in the format: service-metric (i.e: cinder-snapshots)
+      --disable-service.network  Disable the network service exporter
+      --disable-service.compute  Disable the compute service exporter
+      --disable-service.image    Disable the image service exporter
+      --disable-service.volume   Disable the volume service exporter
+      --disable-service.identity
+                                 Disable the identity service exporter
+      --disable-service.object-store
+                                 Disable the object-store service exporter
+      --disable-service.load-balancer
+                                 Disable the load-balancer service exporter
+      --disable-service.container-infra
+                                 Disable the container-infra service exporter
+      --disable-service.dns      Disable the dns service exporter
+      --disable-service.baremetal
+                                 Disable the baremetal service exporter
+      --disable-service.gnocchi  Disable the gnocchi service
       --version                  Show application version.
+
+Args:
+  <cloud>  name or id of the cloud to gather metrics from
 ```
 
 ### Scrape options
 
+In legacy mode cloud and metrics to be scraped are specified as argument or flags as described above.
+To select multi cloud more the -all-cloud flag need to be used
+In that case metrics and clouds are specified in the http scrape request as described below.
 Which cloud (name or id from the `clouds.yaml` file) or what services from the cloud to scrape, can be specified as the parameters to http scrape request.
 
 Query Parameter | Description
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Other options as the binding address/port can by explored with the --help flag.
 The exporter can operate in 2 modes
 - A Legacy mode (targetting one cloud) in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
 - A multi cloud mode in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL.
+  And where `/metrics` URL is serving own exporter metrics 
 
 You can build it by yourself by cloning this repository and run:
 
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ make common-build
 ```
 Multi cloud mode
 ```sh
-./openstack-exporter --os-client-config /etc/openstack/clouds.yaml --all-clouds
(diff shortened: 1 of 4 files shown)


niedbalski:
Hello @rustycl0ck 

Thank you so much. I think this is a great feature, however, I have some observations prior to review this WIP:

1. I would like to avoid breaking compatibility with existing configurations, therefore, removing the "cloud" positional argument shouldn't be the way to go, instead, I propose to make that parameter multiple and optional instead of mandatory.  There is also
a chance that some people wants to expose multiple clouds but not all of them (for security or other concerns), so additionally, a new parameter (--all-clouds or alike) should be present (and ignore the clouds positional argument if set).

./openstack-exporter --config clouds.yaml 1.cloud.io 2.cloud.io  

and

./openstack-exporter --config clouds.yaml --all-clouds

For picking up the default one, we could implement another parameter --default-cloud that if not set, first passed should be the default. 

./openstack-exporter --config clouds.yaml --default-cloud 1.cloud.io 2.cloud.io  3.cloud.io [...]


This default-cloud should be the one displayed in /metrics 

2. Changing the endpoint from /metrics to /probe is also breaking compatibility, therefore, I suggest 
to keep using the /metrics endpoint with the additional arguments. If the "cloud" argument(s) have been passed
the /metrics endpoint should behave as it is now (using the default provided cloud), if the argument cloud?= is passed
then scrape those metrics out. 

3. The same should apply for --disable-metric, I don't oppose to the approach of passing them through the query string,
but I'd like to leave the original behavior enabled and apply across all the clouds (and perhaps demote it in a further release)

I am really looking forward for your input on this.

niedbalski:
@rustycl0ck hi!, any feedback on this?

rustycl0ck:
@niedbalski Making the `cloud` argument optional (for backward compatibility) and also for providing access to selected clouds, makes sense. In addition to the `--default-cloud` this could work well. So if I understand correctly: either one or more cloud names must be passed or the `--default-cloud` should be set on CLI.

Regarding point 2, I think at some point in future, the `/metrics` endpoint (for serving the application metrics of the `openstack-exporter` itself) and the `/probe` endpoint (for serving the scraped metrics from the external service/cloud) have to be separated. For now, to keep it backward compatible, we can serve the cloud metrics at `/probe` by default but also keep the option of serving them at `/metrics` if explicitly passed from command line. This would provide users the flexibility to choose if they want to go with the default `/probe` endpoint and make corresponding changes to their prometheus, or to not make any change to prometheus and preserve the old behavior by passing an extra command line parameter.

Regarding point 3, I didn't really try to keep it backward compatible because I had already broken that in points 1 and 2 above. But now that we are aiming to keep it backward compatible, we can choose keep this behavior as it is and remove it in some future release when comfortable.

Sorry for the late reply. But I don't think I'll be able to spend more time on this in the immediate future. I'll request @ichtar to take this further too, since they seem to be more active on it recently.

@ichtar Please drop an acknowledgement comment if you think you will have the time to take this up from here.

ichtar:
Hi @rustycl0ck and @niedbalski 

yes I have worked on a solution that will brought back backward compatibility for this PR (this is currently a PR on rustyclock PR)
My idea is to provide /probe and /metrics and to be able to toggle to new mode with a --all-clouds flags.
Without this parameter /metrics still behave as before and so will --disable-metrics and mandatory cloud specification
If --all-clouds flag  is set then /probe is used, and disable metrics/clouds are read from query string.
I am not sure it worth now to add more complexity to exclude clouds from config file as one could create multiple config files based on different needs. --default-cloud is not needed because if you use legacy mode you have to specify the cloud you want to interact with.

Let me know what you think @niedbalski

niedbalski:
@ichtar 

Overall looks good to me. Feel free to update this PR so I can review it.

niedbalski:
Hello @rustycl0ck @ichtar 

Can you check the failing lint tests? https://app.circleci.com/pipelines/github/openstack-exporter/openstack-exporter/307/workflows/e07163b5-638f-42db-b2f7-e90fcf395924/jobs/940

niedbalski:
@ichtar @rustycl0ck hello!

The change looks overall good to me, I am willing to merge it, but I am still not sold on the idea
of having 2 different endpoints /metrics and /probe for pretty much the same purpose (except for the query string filtering).

Why not keeping a single handler /metrics that accepts query strings and that by default it behaves as today when
no argument is given? Is there a fundamental reason on why you think this should be split? 

I look forward for your comments.

README.md
```diff
@@ -88,27 +90,30 @@ Flags:
...
-      --disable-service.object-store
-                                 Disable the object-store service exporter
-      --disable-service.load-balancer
-                                 Disable the load-balancer service exporter
-      --disable-service.container-infra
-                                 Disable the container-infra service exporter
-      --disable-service.dns      Disable the dns service exporter
-      --disable-service.baremetal
-                                 Disable the baremetal service exporter
-      --disable-service.gnocchi  Disable the gnocchi service
```
niedbalski, line 106: why was this removed?

ichtar: This was a mistake not to re-add it when backward compatibility has been added back.
it will be corrected

ichtar:
Regarding separated /metrics and /probe it was decided based on prometheus documentation
https://prometheus.io/docs/guides/multi-target-exporter/
where an exporter been use in a multi target fashion is using /probe
`Basic querying of multi-target exporters

There are two ways of querying:

    Querying the exporter itself. It has its own metrics, usually available at /metrics.
    Querying the exporter to scrape another target. Usually available at a "descriptive" endpoint, e.g. /probe. This is likely what you are primarily interested in, when using multi-target exporters.
`

README.md
niedbalski: OK

README.md
```diff
@@ -53,20 +53,22 @@ and must by specified with the `--os-client-config` flag.
 
 Other options as the binding address/port can by explored with the --help flag.
 
-By default the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+By default the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL.
```
niedbalski, line 56: OK, please specify the difference here between /metrics and /probe and how to use both perhaps would be good to point out the prometheus official documentation as a reference for /probe.

niedbalski:
@ichtar @rustycl0ck hello guys, is this ready for review? I'd like to include this in the next release? wonder if you can improve the readme files as indicated there?  thank you!

ichtar:
@niedbalski I have amended documentation and pushed a fix for broken time collection that was unhappy with pedanticRegistry
It is ready for review

README.md
```diff
@@ -53,20 +53,25 @@ and must by specified with the `--os-client-config` flag.
 
 Other options as the binding address/port can by explored with the --help flag.
 
-By default the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+The exporter can operate in 2 modes
+- A Legacy mode (targetting one cloud) in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+- A multi cloud mode in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL.
+  And where `/metrics` URL is serving own exporter metrics
```
niedbalski, line 59: +The exporter can operate in 2 modes:
+
+* A Legacy mode (targeting one cloud) in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+* A multi cloud mode in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL and where `/metrics` URL is serving own exporter metrics

README.md
```diff
@@ -53,20 +53,25 @@ and must by specified with the `--os-client-config` flag.
...
 
-By default the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+The exporter can operate in 2 modes
+- A Legacy mode (targetting one cloud) in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/metrics` URL.
+- A multi cloud mode in where the openstack\_exporter serves on port `0.0.0.0:9180` at the `/probe` URL.
+  And where `/metrics` URL is serving own exporter metrics 
 
 You can build it by yourself by cloning this repository and run:
 
 ```sh
```
niedbalski, line 63:
```sh
 make common-build
-./openstack-exporter --os-client-config /etc/openstack/clouds.yaml
+```
+Multi cloud mode
+```sh
+./openstack-exporter --os-client-config /etc/openstack/clouds.yaml --all-clouds
 curl "http://localhost:9180/probe?cloud=region.mycludprovider.org"
 ```
+or  Legacy mode
+
+```sh
+./openstack-exporter --os-client-config /etc/openstack/clouds.yaml myregion.cloud.org
+curl "http://localhost:9180/metrics"
+```

File: README.md [hunk 5]
```diff
@@ -103,14 +110,38 @@ Flags:
...
       --version                  Show application version.
 
 Args:
   <cloud>  name or id of the cloud to gather metrics from
 ```
 
+### Scrape options
+
+In legacy mode cloud and metrics to be scraped are specified as argument or flags as described above.
+To select multi cloud more the -all-cloud flag need to be used
```
niedbalski, line 123: +To select multi cloud more the --all-clouds flag need to be used

exporters/exporter.go
```diff
@@ -91,19 +90,6 @@ func (exporter *BaseOpenStackExporter) Describe(ch chan<- *prometheus.Desc) {
...
-		"openstack_metric":  metricName}
-	metric := prometheus.NewDesc(
-		"openstack_metric_collect_seconds",
-		"Time needed to collect metric from OpenStack API",
-		nil,
-		metricPromatheuslabels)
-	log.Debugf("Adding metric for collecting timings: %+s", metric)
-	ch <- prometheus.MustNewConstMetric(metric, prometheus.GaugeValue, collectTimeSeconds)
-}
-
```
niedbalski, line 106: Removing the specific labels, will make this fail with:


INFO[0007] Adding metric: limits_memory_used to exporter: nova  source="exporter.go:164"
2020/12/03 10:09:39 http: panic serving [::1]:60370: descriptor Desc{fqName: "openstack_metric_collect_seconds", help: "Time needed to collect metric from OpenStack API", constLabels: {}, variableLabels: [openstack_service openstack_metric]} already exists with the same fully-qualified name and const label values
goroutine 36 [running]:
net/http.(*conn).serve.func1(0xc0001a4820)
        /snap/go/current/src/net/http/server.go:1801 +0x147
panic(0x95a520, 0xc000431d10)
        /snap/go/current/src/runtime/panic.go:975 +0x47a
github.com/prometheus/client_golang/prometheus.(*Registry).MustRegister(0xc0000b0050, 0xc000229b10, 0x1, 0x1)
        /home/niedbalski/go/pkg/mod/github.com/prometheus/client_golang@v1.2.1/prometheus/registry.go:400 +0xad
main.probeHandler(0xab0a60, 0xc0000ca000, 0xc0001ce500)
        /home/niedbalski/GolandProjects/openstack-exporter/main.go:120 +0x6e5
net/http.HandlerFunc.ServeHTTP(0xa30530, 0xab0a60, 0xc0000ca000, 0xc0001ce500)
        /snap/go/current/src/net/http/server.go:2042 +0x44
net/http.(*ServeMux).ServeHTTP(0xdd2840, 0xab0a60, 0xc0000ca000, 0xc0001ce500)
        /snap/go/current/src/net/http/server.go:2417 +0x1ad
net/http.serverHandler.ServeHTTP(0xc0001e2000, 0xab0a60, 0xc0000ca000, 0xc0001ce500)
        /snap/go/current/src/net/http/server.go:2843 +0xa3
net/http.(*conn).serve(0xc0001a4820, 0xab2ca0, 0xc0001a0540)
        /snap/go/current/src/net/http/server.go:1925 +0x8ad
created by net/http.(*Server).Serve
        /snap/go/current/src/net/http/server.go:2969 +0x36c


Please revert the change or make sure that labels are unique for this metric.

main.go
```diff
@@ -88,7 +75,105 @@ func main() {
 			log.Error(err)
 		}
 	})
+	if *allClouds {
+		http.HandleFunc("/probe", probeHandler)
+		http.Handle("/metrics", promhttp.Handler())
```
niedbalski, line 80: Please add a log entry that specifies that is running in probe mode and which clouds were loaded.

main.go
```diff
@@ -88,7 +75,105 @@ func main() {
...
+	} else {
+		http.HandleFunc(*metrics, metricHandler(services))
```
niedbalski, line 82: Please add a log entry that specifies that is running in legacy mode

niedbalski (changes requested):
Hello @rustycl0ck @ichtar 

This change looks like a great addition to the exporter. Thank you. I left some comments
in the review that I'd like you to address before merging. Please review and fix.

exporters/exporter.go
ichtar: Fixed that, I was doing unit testing on only 1 exporter for speed sake ...

main.go
ichtar: Added log message
Regarding displaying list of clouds I don't think it worth doing it. Trade off between adding meaningful information vs adding more code looks not beneficial.
If you agree I would rather not doing it

main.go
ichtar: Done

README.md
ichtar: fixed

README.md
ichtar: fixed

README.md
```diff
@@ -87,6 +100,8 @@ Flags:
       --prefix="openstack"       Prefix for metrics
       --endpoint-type="public"   openstack endpoint type to use (i.e: public, internal, admin)
       --collect-metric-time      time spent collecting each metric
+      --disable-slow-metrics     disable slow metrics for performance reasons
+      --all-clouds               Toggle the multiple cloud scrapping mode under /probe?cloud=
```
frittentheke, line 104: If you don't mind my comment on this PR ... why not call the parameter "--multi-cloud" instead of "--all-clouds".
`all` somehow implies this causes all clouds to be returned in contract to a default. But that is not the case, it's just enabling support for "multi-cloud".

(Nitpick: it's `scraping`, not `scrapping`)

niedbalski: @ichtar Apologies for not debating this earlier, but I tend to agree with @frittentheke's comment here. --multi-cloud sounds reasonable considering that it's the opposite to single cloud. I considered calling it --all-clouds as the verb/action of loading all clouds from the configuration, but multi-cloud encapsulates the feature's essence better. 

Also, the nitpick is a good finding,

niedbalski:
LGTM @rustycl0ck @ichtar minor nitpicks noted by @frittentheke and we're ready to merge.

README.md
ichtar: changed scrapping to scraping and all-clouds to multi-cloud

niedbalski:
LGTM, thanks for this patch!

Issue #144, opened 2020-11-26
No way to bind a exporter to a tcp4 address


isaac-gardner:
When running exporter with ipv4 address "0.0.0.0:1234" exporter still defaults to ipv6.

The `http.ListenAndServe` is provided by convenience only and a tcp listener should be used instead.

Current workaround is to disable ipv6.

PR #147, opened 2020-12-06
Bind a exporter to only needed tcp version

alexeymyltsev:
This PR bind exporter only to needed tcp version. 
If ip did not provided exporter will listen on all tcp versions.
It should solve https://github.com/openstack-exporter/openstack-exporter/issues/144

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,8 +2,11 @@ package main
 
 import (
 	"fmt"
+	"net"
 	"net/http"
 	"os"
+	"regexp"
+	"strings"
 
 	"github.com/openstack-exporter/openstack-exporter/exporters"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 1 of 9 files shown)


niedbalski:
@alexeymyltsev Thank you for looking into this!. please rebase your patch and check the failing tests.

main.go
```diff
@@ -63,7 +65,28 @@ func main() {
...
+	}
+
+	if enabledExporters == 0 {
+		log.Errorln("No exporter has been enabled, exiting")
+		os.Exit(-1)
+	}
+
+	sm := http.NewServeMux()
+	sm.Handle(*metrics, promhttp.Handler())
+	sm.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
```
niedbalski, line 89: why is this change required? lines 68 to 89

main.go
```diff
@@ -84,8 +107,31 @@ func main() {
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	log.Infoln("Starting HTTP server on", *bind)
-	log.Fatal(http.ListenAndServe(*bind, nil))
+	log.Fatal(http.Serve(l, sm))
+}
+
+func ip4or6(s string) string {
```
niedbalski, line 121: Would be possible to move this function into a utils.go file or module alongside with removeElements?

niedbalski (changes requested):
Hello @alexeymyltsev thank you for this fix. left a minor set of nitpicks as code reviews.

main.go
alexeymyltsev: You are right, i didn't dig enough to http lib.
Now it should be better.

main.go
alexeymyltsev: done.
Please fix comment for exporters.RemoveElements if it is not clear description.

niedbalski:
LGTM

PR #150, opened 2020-12-14
fixes #122

### LINKED ISSUE #122: [bug] metric:openstack_neutron_up is not correct
when I stop or run neutron_server, the openstack_neutron_up is always 0

niedbalski:
check if all metrics are down, if so, mark the service as down.


niedbalski:
LGTM

PR #159, opened 2021-02-05
Add UpdateAt label to nova agent_state metrics

lrainsun:
When nova cell enabled, we have two nova-conductor service, one is
nova-conductor and another is super nova-super-conductor, they have same
name in different databases (nova_cell0 & nova_cell1).
The only difference for the agent_state metrics is the service ID, but
service ID is actually a db seqNo, they can be same in different
tables. In our cases, we met two nova-conductors have the same label,
that will cause metrics failed to store. So add UpdateAt to identify
different nova-conductor services.

### DIFF
--- a/exporters/nova.go
+++ b/exporters/nova.go
@@ -63,7 +63,7 @@ var defaultNovaMetrics = []Metric{
 	{Name: "availability_zones", Fn: ListAZs},
 	{Name: "security_groups", Fn: ListComputeSecGroups},
 	{Name: "total_vms", Fn: ListAllServers},
-	{Name: "agent_state", Labels: []string{"id", "hostname", "service", "adminState", "zone", "disabledReason"}, Fn: ListNovaAgentState},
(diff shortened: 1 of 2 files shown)


niedbalski:
Hello @lrainsun 

Thank you very much for this contribution. Could you please check the failing tests before I review?

alexeymyltsev:
Hello, I worry about numbers of metrics, each new label will generate new metric and it could be not so good for monitoring systems.

lrainsun:
> Hello @lrainsun
> 
> Thank you very much for this contribution. Could you please check the failing tests before I review?

of course, trying to fix the test

lrainsun:
> Hello, I worry about numbers of metrics, each new label will generate new metric and it could be not so good for monitoring systems.

Yes, you're right, I'm trying another fix for this problem
https://github.com/openstack-exporter/openstack-exporter/pull/160

Issue #170, opened 2021-03-17
Slow metrics for volume status


alexeymyltsev:
We should exclude `volume_status` from slow metrics, or separate `volumes` and `volume_status` on two different functions. 
For this moment if use `--disable-slow-metrics` we will get panic.
```anic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x14cf476]

goroutine 97 [running]:
github.com/openstack-exporter/openstack-exporter/exporters.ListVolumes(0xc00032dbc0, 0xc00058f080, 0x19f3ac0, 0x2)
        /Users/amyltsev/Documents/Leroy Merlin/github/openstack-exporter/exporters/cinder.go:111 +0x1176
github.com/openstack-exporter/openstack-exporter/exporters.(*BaseOpenStackExporter).RunCollection(0xc00032dbc0, 0xc00041ddb0, 0x161bbf4, 0x7, 0xc00058f080, 0x0, 0x0)
        /Users/amyltsev/Documents/Leroy Merlin/github/openstack-exporter/exporters/exporter.go:96 +0x1ff
github.com/openstack-exporter/openstack-exporter/exporters.(*BaseOpenStackExporter).Collect(0xc00032dbc0, 0xc00058f080)
        /Users/amyltsev/Documents/Leroy Merlin/github/openstack-exporter/exporters/exporter.go:117 +0x118
github.com/prometheus/client_golang/prometheus.(*Registry).Gather.func1()
        /Users/amyltsev/go/pkg/mod/github.com/prometheus/client_golang@v1.2.1/prometheus/registry.go:443 +0x1a2
created by github.com/prometheus/client_golang/prometheus.(*Registry).Gather
        /Users/amyltsev/go/pkg/mod/github.com/prometheus/client_golang@v1.2.1/prometheus/registry.go:535 +0xe8e
exit status 2```

PR #171, opened 2021-03-17
exclude volume_status from slow metrics

alexeymyltsev:
https://github.com/openstack-exporter/openstack-exporter/issues/170

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ Which cloud (name or id from the `clouds.yaml` file) or what services from the c
 Query Parameter | Description
 --- | ---
 `cloud` | Name or id of the cloud to gather metrics from (as specified in the `clouds.yaml`)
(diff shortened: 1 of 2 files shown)


niedbalski:
@alexeymyltsev What's the purpose of this change? Why should we remove volume_status from slow metrics?

niedbalski:
Apologies @alexeymyltsev , that's documented on #170

niedbalski:
@alexeymyltsev CI tests are failing, also .. why not leaving the volume_status in a different function?

alexeymyltsev:
@niedbalski failed build, same fail what was before `sudo snap install go --classic` .
I do not want add separate function because I do not want duplicated calls to API. 
I see it as tradeoff between disabling metrics with one parameter and additional API calls.

Issue #180, opened 2021-04-20
[Feature Request] change/add Metric openstack_cinder_volume


brolga:
​​​​​The cinder volume metric ist exported as such:

```
openstack_cinder_volume_status
{
Region="RegionOne",
bootable="true",
id="173f7b48-c4c1-70-9acc-086b39073506",
name="test-volume",
size="1",
status="available",
tenant_id="bab7d5c60cd041a0a36f7c4b6e1dd978",
volume_type="lvmdriver-1",
server_id="f4fda93b-06e0-4743-8117-bc8bcecd651b"	
}
4.0 (float)
```

I propose the following change to this metric:

**Encode the status inside a label and not as a metric value.
Encode the volume size not as a Label but in the metric value.
Rename the metric to openstack_cinder_volume** 

This would enable the following prometheus queries, that are not easily possible with the current metric:

-  get the amount of volumestorage with a certain state
sum(openstack_cinder_volume{status="maintenance})

- get the count of volumes by a certain status and display status in cleartext
count(openstack_cinder_volume)by(status)

- sum up the amount volumestorage by tenant and volumetype
sum(openstack_cinder_volume)by(tenant_id,volume_type)

-  get the amount of volumes in errorstate
count(openstack_cinder_volume(status=~"error.*"}

Imho the ability to use this kind of queries would by very handy for resource management dashboards and also prometheus alarming.

alexeymyltsev:
Hi @brolga ,
Your suggestion looks reasonable and I have created PR for it but I worry about people who already use openstack_cinder_volume(like me).
Additionally I changed name to volume_gb, it should be self explanatory for value.

@niedbalski what do you think about this change?

brolga:
The suffix `_gb` is perfect for me.

A new Feature in 2 hours that solves my need ;-)

Very impresive, thank you very  much !

niedbalski:
@alexeymyltsev @brolga I think the addition is very reasonable and I am in favor of the change. However, I dislike very much the idea of deprecating a metric without further notice in a new release.

I will post comments in pr #181 on how I think we could improve this change to make it have a minor impact for
the users upgrading.

brolga:
I think it es reasonable add a new metric name and keep the original one laso in place.

PR #181, opened 2021-04-20
[wip] change volume_status to volume_gb

alexeymyltsev:
https://github.com/openstack-exporter/openstack-exporter/issues/180


niedbalski:
@alexeymyltsev thanks for this patch.

I think we'll need to create a deprecation cadence process for metrics that extends the metrics
for a X+1 release to avoid disruptive changes to land without further notice, which includes a few things:

1. A new README.md section named "deprecated", which has (metric name | last release | deprecation on | notes), so as an example
this metric will be last supported in 1.4 | 1.5 won't have it | deprecated in favor of volume_gb.
2. A --skip-deprecated flag that will avoid deprecated metrics, extend the default metric (deprecated = false) that will show a log.warn message saying this metric has been deprecated and will be removed as of release X.Y 
3. Extend the default metric to include a deprecated{version, bool} (nil, false) by default.

@alexeymyltsev ideas/opinions?

alexeymyltsev:
@niedbalski your plan looks good for me. 
This process should make exports more mature.

Should we implement it in this PR or will be added new one?

niedbalski:
@alexeymyltsev are you planning to work on it? Its fine if you want to include that in this PR. (no objections on my end)

alexeymyltsev:
@niedbalski ok, I will work on it.

exporters/cinder.go
```diff
@@ -21,43 +20,11 @@ type CinderExporter struct {
...
-	}
-	return -1
-}
-
 var defaultCinderMetrics = []Metric{
 	{Name: "volumes", Fn: ListVolumes},
 	{Name: "snapshots", Fn: ListSnapshots},
 	{Name: "agent_state", Labels: []string{"uuid", "hostname", "service", "adminState", "zone", "disabledReason"}, Fn: ListCinderAgentState},
-	{Name: "volume_status", Labels: []string{"id", "name", "status", "bootable", "tenant_id", "size", "volume_type", "server_id"}, Fn: nil},
+	{Name: "volume_gb", Labels: []string{"id", "name", "status", "bootable", "tenant_id", "size", "volume_type", "server_id"}, Fn: nil},
```
brolga, line 27: I think the label "size" ist not neede wens size is already in the metric.

alexeymyltsev: you are right. 
Below commit remove it.

brolga:
The is another thing that comes into my mind. 
Please bear with me.

 Prometheus says on its webpage
```Metrics must use base units (e.g. seconds, bytes) and leave converting them to something more readable to graphing tools.```

**Comment by alexeymyltsev** [comment]
@brolga 
> Prometheus says on its webpage
Metrics must use base units (e.g. seconds, bytes) and leave converting them to something more readable to graphing tools.

you mean we should use bytes instead gb?
Unfortunately OpenStack for volumes size return value in GB and all another metrics already have suffix _gb.
I think we can skip this requirements from Prometheus.

**Comment by brolga** [comment]
This is ok for me.

**Comment by niedbalski** [comment]
Thanks! I am going to set this review as work in progress @alexeymyltsev , until we have the deprecated metrics support.

**Review by niedbalski — CHANGES_REQUESTED** [review]
On hold.

File: exporters/cinder.go [hunk 2]
```diff
@@ -105,16 +109,31 @@ func ListVolumes(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metric) e
...
-				prometheus.GaugeValue, float64(mapVolumeStatus(volume.Status)), volume.ID, volume.Name,
-				volume.Status, volume.Bootable, volume.TenantID, strconv.Itoa(volume.Size), volume.VolumeType, "")
+			ch <- prometheus.MustNewConstMetric(exporter.Metrics["volume_gb"].Metric,
+				prometheus.GaugeValue, float64(volume.Size), volume.ID, volume.Name,
+				volume.Status, volume.Bootable, volume.TenantID, volume.VolumeType, "")
+		}
+	}
+
+	// Volume status metrics
+	if !exporter.ExporterConfig.DisableDeprecatedMetrics {
```
niedbalski, line 126: I think the correct way for this would be to move this metric into a separated function and simply let the if exporter.isDeprecatedMetric to check for this condition.

README.md
```diff
@@ -89,37 +89,38 @@ The current list of command line options (by running --help)
...
+      --disable-service.volume         Disable the volume service exporter
       --disable-service.identity
-                                 Disable the identity service exporter
+                                       Disable the identity service exporter
       --disable-service.object-store
-                                 Disable the object-store service exporter
+                                       Disable the object-store service exporter
       --disable-service.load-balancer
-                                 Disable the load-balancer service exporter
+                                       Disable the load-balancer service exporter
```
niedbalski, line 116: why the trailing whitespaces?

niedbalski, line 116: why the trailing whitespaces?

exporters/cinder.go
```diff
@@ -57,7 +57,8 @@ var defaultCinderMetrics = []Metric{
 	{Name: "volumes", Fn: ListVolumes},
 	{Name: "snapshots", Fn: ListSnapshots},
 	{Name: "agent_state", Labels: []string{"uuid", "hostname", "service", "adminState", "zone", "disabledReason"}, Fn: ListCinderAgentState},
-	{Name: "volume_status", Labels: []string{"id", "name", "status", "bootable", "tenant_id", "size", "volume_type", "server_id"}, Fn: nil},
+	{Name: "volume_gb", Labels: []string{"id", "name", "status", "bootable", "tenant_id", "volume_type", "server_id"}, Fn: nil},
+	{Name: "volume_status", Labels: []string{"id", "name", "status", "bootable", "tenant_id", "size", "volume_type", "server_id"}, Fn: nil, Slow: false, DeprecatedVersion: "1.5"},
```
niedbalski, line 61: 1.4, 1.5 we will remove it :-)

niedbalski (changes requested):
Hey @alexeymyltsev great work on this. I left a couple of comments to be addressed before merging.

niedbalski:
@alexeymyltsev added a couple of comments. I'd like to release 1.4.0 with this patch included , so holding on this.

README.md
alexeymyltsev: I think view like this 
image could be better.
Should I remove this changes?

exporters/cinder.go
alexeymyltsev: Ok I will add in warning that metric will be removed in next release

alexeymyltsev:
I have added separate function for volume_status.

alexeymyltsev:
@niedbalski is there some additional comments for this PR?

niedbalski (approved):
LGTM

PR #183, opened 2021-04-29
add project_id to openstack_container_infra_cluster_status

skunkie:
project_id helps make dashboards with a filter based on projects. It is present in other metrics, so it is reasonable to have it here as well.


niedbalski:
LGTM , thanks a lot @skunkie for this contribution. Adding you to the CONTRIBUTORS.md list, thanks!

Issue #194, opened 2021-06-04
malformed listen address causes panic


papaburgs:
logs from journalctl
```
Jun 04 14:44:56 host openstack-exporter[25299]: panic: runtime error: index out of range [0] with length 0
Jun 04 14:44:56 host openstack-exporter[25299]: goroutine 1 [running]:
Jun 04 14:44:56 host openstack-exporter[25299]: github.com/openstack-exporter/openstack-exporter/exporters.IP4or6(0x7ffd37996df8, 0x7, 0x1, 0xb24a40)
Jun 04 14:44:56 host openstack-exporter[25299]: /app/exporters/utils.go:193 +0x1a7
Jun 04 14:44:56 host openstack-exporter[25299]: main.main()
Jun 04 14:44:56 host openstack-exporter[25299]: /app/main.go:89 +0x84c
```

I tried running with an invalid listen address (a problem with quoting)

Would expect it to fail, but at least log what was incorrect

alexeymyltsev:
Hello @papaburgs.
Thank you for your report. 
Could you share example for problem reproduction?

papaburgs:
Anything that doesn't match the regex will cause this, for example:
```
./openstack-exporter --multi-cloud --web.listen-address=""
```
or using "9888". 

I think the re.FindAllStrings is returning nil in that case, which causes the panic when it tries to be refrerenced.
I wouldn't expect it it succeed with an empty string, but at least log what the issue is.

PR #195, opened 2021-06-06
Fix IP4or6 if wrong port

alexeymyltsev:
This PR should fix https://github.com/openstack-exporter/openstack-exporter/issues/194


main.go
```diff
@@ -86,6 +86,10 @@ func main() {
 		http.HandleFunc(*metrics, metricHandler(services))
 	}
 
+	if *bind == "" {
+		log.Fatal("--web.listen-address is empty. Please specify address:port to listen on.")
```
niedbalski, line 90: Should be a default value, rather than a log fatal

alexeymyltsev: last commit should fix it

niedbalski:
LGTM

PR #207, opened 2021-10-19
Escaping pipe characters in tables of readme

cuttingedge1109:
`|` pipe in the value is treated as column deliminator. To use `|` as string in the table, need to use HTML character code `&#124;`.


niedbalski:
Sorry, But I don't see why this would be preferred/required.

PR #209, opened 2021-11-09
Add parent_id field to keystone project_info

priteau:
This can be useful in deployments using hierarchical multitenancy (nested projects).


niedbalski:
Thanks @priteau, merged.

Issue #213, opened 2021-12-08
Unable to get cinder metrics


sigtryggr:
During the connecting to Openstack API, openstack-exporter got such an error:

{"log":"time=\"2021-12-08T20:27:31+03:00\" level=error msg=\"enabling exporter for service volumev3 failed: unable to create a service client for volumev3\" source=\"main.go:169\"\n","stream":"stderr","time":"2021-12-08T17:27:31.829477295Z"}

Meanwhile, in th network dump we can see, that API service returns volumev3 endpoint correctly:

{
            "endpoints":[
               {
                  "id":"112b312a233b44cfbfc6ca2943a824c5",
                  "interface":"public",
                  "region_id":"RegionOne",
                  "url":"http://10.1.38.113:8776/v3/f8e6a2a1027245b4b65f6458299a8e2f",
                  "region":"RegionOne"
               },
               {
                  "id":"bf793a17deeb4e4783e42da012bc2fc0",
                  "interface":"internal",
                  "region_id":"RegionOne",
                  "url":"http://10.1.38.113:8776/v3/f8e6a2a1027245b4b65f6458299a8e2f",
                  "region":"RegionOne"
               },
               {
                  "id":"f5aff43611bf47148c15923dc1375663",
                  "interface":"admin",
                  "region_id":"RegionOne",
                  "url":"http://10.1.38.113:8776/v3/f8e6a2a1027245b4b65f6458299a8e2f",
                  "region":"RegionOne"
               }
            ],
            "id":"fdbd5f46d5e94deaa53a610b0c433aa6",
            "type":"volumev3",
            "name":"cinderv3"
         }

What could be a cause of such behaviour?

sigtryggr:
> Hi @sigtryggr, Could you show attributes and config for exporter?

openstack-exporter was deployed with kolla-ansible, so, there are no special configuration specified. There is a command:

/etc/openstack/openstack-exporter --endpoint-type=internal --os-client-config=/etc/openstack/clouds.yml --web.listen-address=10.1.38.131:9198 default --disable-service.dns --disable-service.object-store --disable-service.load-balancer --disable-metric=neutron-loadbalancers --disable-metric=neutron-loadbalancers_not_active, --log.level=debug"

There is clouds.yml:

clouds:
 default:
   region_name: RegionOne
   identity_api_version: 3
   identity_interface: internal
   auth:
     username: admin
[...]

PR #215, opened 2021-12-22
Update gophercloud and default to Cinder API v3

osfrickler:
This resolves an issue when trying to access the Cinder API v3 endpoint.

The call signature for hypervisors.List() has changed, update our
invocation accordingly.

Closes: #213


mnaser (approved):
lgtm

PR #217, opened 2022-01-30
Added availability_zone to openstack_neutron_agent_state

stelterlab:
We have openstack victoria environment with neutron network agents which are losing their availability zone from time to time. So new virtual machines have no connectivity. We would like to be able monitor this.

So I added a new label to the existing metric openstack_neutron_agent_state.


alexeymyltsev:
Could you update README regarding new labels.

stelterlab:
ACK. I updated the README now.

PR #218, opened 2022-02-01
Update cinder

vdlobanov:


niedbalski (changes requested):
Could you add a test covering your change?

PR #219, opened 2022-02-03
New cinder exporter metrics

vdlobanov:
added new cinder-exporter metrics:
volume_type_used_gb
snapshots_count_by_volume_type
snapshots_count


exporters/cinder.go
```diff
@@ -308,7 +312,52 @@ func ListVolumeLimits(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metr
 		if err != nil {
 			return err
 		}
-
+		jsonLimits := quotasets.GetUsage(exporter.Client, p.ID).PrettyPrintJSON()
```
niedbalski, line 315: Can you move this entire logic into its own unit/method so you can add a test for this? I think it is too much logic for a single function.

exporters/cinder.go
```diff
@@ -308,7 +312,52 @@ func ListVolumeLimits(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metr
...
+		var jsonParse map[string]interface{}
+		_ = json.Unmarshal([]byte(jsonLimits), &jsonParse)
+		quota := jsonParse["quota_set"].(map[string]interface{})
```
niedbalski, line 318: quota, ok := 
if !ok { return fmt.Errorf() }

exporters/cinder.go
```diff
@@ -308,7 +312,52 @@ func ListVolumeLimits(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metr
...
+				if words[0] == "gigabytes" {
+					if wordsLen == 2 {
+						disksQuota := quota[i].(map[string]interface{})
+						inUseGb := disksQuota["in_use"].(float64)
+						volType := words[1]
+						ch <- prometheus.MustNewConstMetric(exporter.Metrics["volume_type_used_gb"].Metric,
+							prometheus.GaugeValue, inUseGb, p.Name, p.ID, volType)
+					} else if wordsLen > 2 {
+						jsonHead := strings.Join(words, "_")
+						disksQuota := quota[jsonHead].(map[string]interface{})
```
niedbalski, line 332: casting may fail and break the exporting

exporters/cinder.go
```diff
@@ -308,7 +312,52 @@ func ListVolumeLimits(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metr
...
+						inUseGb := disksQuota["in_use"].(float64)
```
niedbalski, line 333: casting may fail and break the exporting

niedbalski (changes requested):
Thanks a lot for your contribution!. Please refactor the method a bit so that we can test this logic independently. Also, generally assert type conversions for failure and check for existence of keys in the dictionaries.

exporters/cinder.go
vdlobanov: Thank you for reviewing!
Moved logic to function in 085a1ae20af7403bcf732e4dc0d2089c129747f3

exporters/cinder.go
vdlobanov: fixed in 085a1ae20af7403bcf732e4dc0d2089c129747f3

exporters/cinder.go
vdlobanov: fixed in 085a1ae20af7403bcf732e4dc0d2089c129747f3

exporters/cinder.go
vdlobanov: fixed in 085a1ae20af7403bcf732e4dc0d2089c129747f3

niedbalski (changes requested):
Hello,

Can you please add unit tests for the function you are adding?

Thanks!

PR #227, opened 2022-02-25
Restrict metrics for a given domain

### LINKED ISSUE #110: [Feature Request] Restrict metrics to a domain
Hi,

We have performances issues with openstack_exporter because it gather metrics from every single components in any domain in Openstack.
In our case, we have plenty of projects create by Heat that are empty but still, the project object remains.
The ration is basically something around 70/1 "fake" projects comapred to real ones.
Those projects are created in a "heat" domain while our legitimate projects are created in the "default" domain.

It would be great to add something like `--domain <domainID>` in the command line to restrict metrics collection to this domain and not all (whoich should remain the default behavior).

I did a small hack by hardcoding the domainID on the master branch and my metrics collection time went down from more than a minute to less than 10s.

Best regards

fydrah:
For services supporting domain options, provide
the DomainID to filter metrics for.

New flag `--domain-id` added.
Default behavior is still to gather metrics for all
domains.

Closes #110

This could improve scrape time if openstack cloud has a lot
of heat stacks for example (~1min less for ~1500 heat stacks).

Currently, this only support Domain **ID** (not Name),  a future improvement
could be to support domain names, a list of domains, and/or a regex filter (but this will imply more API calls..)

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ Flags:
       --disable-cinder-agent-uuid  
                                  Disable UUID generation for Cinder agents
       --multi-cloud              Toggle the multiple cloud scraping mode under /probe?cloud=
+      --domain-id=DOMAIN-ID      Gather metrics only for the given Domain ID (defaults to all domains)
       --disable-service.network  Disable the network service exporter
       --disable-service.compute  Disable the compute service exporter
       --disable-service.image    Disable the image service exporter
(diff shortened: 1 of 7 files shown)


Hybrid512 (approved):
As stated, would fix #110 by restricting metrics to 1 domain only but maintain default behavior in case this feature is not used or wanted.

niedbalski:
@priteau  do you have bandwidth to review this pr?

priteau:
It would be useful to document which metrics this option would impact.

fydrah:
I added a new section in the README, and I also rebased the branch on latest master commit (and removed the merge commit on the branch)

fydrah:
Hi @priteau! Did you have some time to check the updates on this PR?

Issue #32, opened 2019-10-03
verify: False but still `x509: certificate signed by unknown authority`


kiblik:
I tried to disable cert verification but exporter still returns `x509: certificate signed by unknown authority`

config:
```yaml
clouds:
 default:
   auth:
     username: xxx
     password: yyy
     project_name: zzz
     verify: False
     auth_url: https://XXX:5000/v2.0
```
Run:
```bash
# docker run -v /opt/os_exporter/:/etc/openstack/ -it quay.io/niedbalski/openstack-exporter-linux-amd64:master default
INFO[0000] Starting openstack exporter version (version=, branch=, revision=) for cloud: default  source="main.go:38"
INFO[0000] Build context (go=go1.11.13, user=, date=)    source="main.go:39"
ERRO[0000] enabling exporter for service network failed: Post https://XXX:5000/v2.0/tokens: x509: certificate signed by unknown authority  source="main.go:51"
ERRO[0000] enabling exporter for service compute failed: Post https://XXX:5000/v2.0/tokens: x509: certificate signed by unknown authority  source="main.go:51"
ERRO[0000] enabling exporter for service image failed: Post https:///XXX:5000/v2.0/tokens: x509: certificate signed by unknown authority  source="main.go:51"
ERRO[0000] enabling exporter for service volume failed: Post https://XXX:5000/v2.0/tokens: x509: certificate signed by unknown authority  source="main.go:51"
ERRO[0000] enabling exporter for service identity failed: Post https://XXX:5000/v2.0/tokens: x509: certificate signed by unknown authority  source="main.go:51"
INFO[0000] Starting HTTP server on :9180                 source="main.go:72"
```

niedbalski:
Hello @kiblik 

Unfortunately, this is a regression introduced by the switch to the gophercloud library. The verify option goes one level up in the configuration YAML (Check README.md), I have fixed it on the code
to pass a new transport to the HTTP client that ignores the SSL verification.

This has to be fixed in the gophercloud library, by the time being, the code #33 should make
this work.

Please let me know if the current master image works for you.

Thanks.

kiblik:
Hi @niedbalski,

You moved verify option one and half level up :smile: because you have _wrong indentation_ there
```bash
$ cat example.yaml 
clouds:
 default:
   region_name: {{ openstack_region_name }}
   identity_api_version: 3
   identity_interface: internal
   auth:
     username: {{ keystone_admin_user }}
[...]

PR #243, opened 2022-07-06
Add ca-trust support from cacert option

Jonher937:
* Add support for including certificates to ca-trust on a per-cloud basis using the `cacert` option in OpenStack clouds config.
* Support for parsing the cacert as path, mentioned as not working in: https://github.com/openstack-exporter/openstack-exporter/issues/32#issuecomment-538015096
* Support for parsing the cacert as PEM encoded data like the README gave as an examle before this PR
* Re-indented the README config to have 2 spaces on each to make it clearer

Tested against a devstack with self-signed certificate.
* If cacert is "", don't call the new `PrepareTLSConfig` function
* If cacert is not empty and containts PEM encoded data, parse it, add to x509 CertPool and use it
* If cacert is not empty and does not start with `---` then expect it to be a path and try to the contents of the path
  * If the data in the file is not valid PEM it will log similar to: `Failed to include additional certificates to ca-trust: Error parsing CA Cert from: ./test.crt`
  * If the file is not there: `Failed to include additional certificates to ca-trust: stat ./test2.crt: no such file or directory`

While passing `cacert` with PEM endoded data does not seem to be in the OpenStack cloud config spec from what I can find, it's a nice addition for when Helm charts lack adding mounts. 
Such as the exporter chart https://github.com/openstack-exporter/helm-charts/pull/11


Jonher937:
Hi @niedbalski could we get a review on this please?

README.md
```diff
@@ -168,25 +168,27 @@ The cloud credentials and identity configuration
...
-     project_name: {{ keystone_admin_project }}
-     project_domain_name: 'Default'
-     project_domain_id: 'Default' // This can replace "project_domain_name"
-     user_domain_name: 'Default'
-     auth_url: {{ admin_protocol }}://{{ kolla_internal_fqdn }}:{{ keystone_admin_port }}/v3
-     cacert: |
-            ---- BEGIN CERTIFICATE ---
+  default:
+    region_name: {{ openstack_region_name }}
+    identity_api_version: 3
```
niedbalski, line 180: why the indenting change? 🤔

exporters/utils.go
```diff
@@ -223,3 +227,33 @@ func RemoveElements(slice []string, drop []string) []string {
 	}
 	return res
 }
+
+func PrepareTLSConfig(caCertFile string) (*tls.Config, error) {
```
niedbalski, line 231: does this function needs to be exported? can you add a unit test? (not blocker, just asking if you could)

exporters/utils.go
```diff
@@ -223,3 +227,33 @@ func RemoveElements(slice []string, drop []string) []string {
...
+	config := &tls.Config{}
+	// Get the SystemCertPool, continue with an empty pool on error
+	trustedCAs, _ := x509.SystemCertPool()
```
niedbalski, line 234: never shadow an error

exporters/utils.go
```diff
@@ -223,3 +227,33 @@ func RemoveElements(slice []string, drop []string) []string {
...
+	}
+	// check if string is not a path, but PEM contents such as: -----BEGIN CERTIFICATE-----
+	if strings.HasPrefix(caCertFile, "---") {
+		ok := trustedCAs.AppendCertsFromPEM(bytes.TrimSpace([]byte(caCertFile)))
+		if !ok {
+			return config, fmt.Errorf("Failed to add cert to trusted roots")
+		}
+	} else {
+		// attempt to read as path
+		_, err := os.Stat(caCertFile)
```
niedbalski, line 246: ReadFile will fail if the file doesn't exist, no need to stat.

niedbalski (changes requested):
Hello @Jonher937 Thanks a lot for this contribution, very useful.

I left you a couple of comments in regards to the code, I would like for you to
address those before I can merge.

Thanks!

README.md
Jonher937: `cacert` was wrongly indented as a start.
Secondly as I wrote in the PR, I did the indentation change so that it's easier to see that cacert is on the same level as auth.
The current README is not consistent on the amount of space for the indentation, for example the `verify` option has 4 spaces, whereas the `auth" has 3 spaces. And those two should be on the same indentation level.

Switching over to 2 spaces for each level shows this difference more clearly.

exporters/utils.go
Jonher937: Doesn't have to be an actual error, it can be a container with only the go binary present and your desired CA chain.

Are you suggesting we add a log entry like so?
```go
	// Get the SystemCertPool, continue with an empty pool on error
	trustedCAs, err := x509.SystemCertPool()
	if trustedCAs == nil {
		log.Infof("Creating a new empty SystemCertPool as we failed to load it from disk: %v", err)
		trustedCAs = x509.NewCertPool()
	}
```

I don't want to treat it as `log.Error` since it's doesn't have to be in a valid use case.

exporters/exporter.go
```diff
@@ -202,6 +202,15 @@ func NewExporter(name, prefix, cloud string, disabledMetrics []string, endpointT
 		log.Infoln("SSL verification disabled on transport")
 		tlsConfig := &tls.Config{InsecureSkipVerify: true}
 		transport = &http.Transport{TLSClientConfig: tlsConfig}
+	} else if config.CACertFile != "" {
```
niedbalski, line 205: just use a if, else if adds unneeded indenting.

Jonher937: That would break functionality for `verify: False` if both verify and `cacert` is set.

niedbalski: > That would break functionality for `verify: False` if both verify and `cacert` is set.

Perhaps, its a bit of refactoring needed here.

just use a if, else if adds unneeded indenting.

```suggestion
var transport http.Transport
var tlsConfig tls.Config

if config.CACertFile != "" {
    //your code should return nil, err in case of not found.
    tlsConfig, err := prepareTLSConfig(config.CACertFile)
    if err != nil {
	log.Errorf("Failed to include additional certificates to ca-trust: %v", err)
    } 
}

if config.Verify != nil && !*config.Verify {
    tlsConfig.InsecureSkipVerify = true
}

transport.TLSClientConfig = tlsConfig
```

Jonher937: As `prepareTLSConfig` would previously return `*tls.Config` I have changed the function name and return type to easier accomodate the requested change.

niedbalski (changes requested):
CI is failing with the following error

```go
Error: exporters/utils_test.go:50:41: ourCertPool.Clone undefined (type *x509.CertPool has no field or method Clone)
Error: exporters/utils_test.go:6[7](https://github.com/openstack-exporter/openstack-exporter/actions/runs/3218945282/jobs/5264362204#step:8:8):18: ourCertPool.Equal undefined (type *x509.CertPool has no field or method Equal)'

```

niedbalski:
> > @Jonher937 CI is failing with the following error
> > ```go
> > Error: exporters/utils_test.go:50:41: ourCertPool.Clone undefined (type *x509.CertPool has no field or method Clone)
> > Error: exporters/utils_test.go:6[7](https://github.com/openstack-exporter/openstack-exporter/actions/runs/3218945282/jobs/5264362204#step:8:8):18: ourCertPool.Equal undefined (type *x509.CertPool has no field or method Equal)'
> > ```
> 
> Go 1.13 is EOL since 11 Aug 2020, would you concider bumping to at least 1.18 or 1.19 which are the versions under support currently?

Sure, send me a PR go get -u ./...  + update to 1.19 and I will approve.

niedbalski:
@Jonher937 Hey, the unit tests are failing spectacularly. I can take a look later, but if you have a cycle 👍🏻

Jonher937:
It's related to us replacing the `http.Transport` in NewExporter after you asked for the refactor of the logical flow.

If you put `verify: false` in `exporters/fixtures/test_config.yaml` and try running tests in main you should see the same failures.

Before we didn't replace the http.Transport in a regular scenario, only if `verify` was set to `false`.
With the current flow the transport would always be replaced and override `httpmock.Activate()` transport.
https://pkg.go.dev/github.com/jarcoal/httpmock#Activate

Jonher937:
@niedbalski Do you have any other preference? Or should I resort back to what I had done initially using `if/else if` to sircumvent this problem?

Since Verify is `default: true`, if we check for Config.Verify first, then secondly act on CAcert if it's present the flow should still act as expected:

* Verify false + CAfile = Verify is disabled, no need to load CAfile
* Verify true(default)  + CAfile = Verify is enabled, we end up on `else if` and load the CAfile

niedbalski:
> @niedbalski Do you have any other preference? Or should I resort back to what I had done initially using `if/else if` to sircumvent this problem?
> 
> Since Verify is `default: true`, if we check for Config.Verify first, then secondly act on CAcert if it's present the flow should still act as expected:
> 
> * Verify false + CAfile = Verify is disabled, no need to load CAfile
> * Verify true(default)  + CAfile = Verify is enabled, we end up on `else if` and load the CAfile

Sounds good to me

Jonher937:
@niedbalski This should be good for an approval

niedbalski (approved):

PR #253, opened 2022-09-17
Glance api up

parsa97:
This PR based on #252 , and will help to get glance availability without get the entire list of images.


niedbalski:
Hello @parsa97 can you review the conflicts and add some unit tests that are specific to this patch?

niedbalski (changes requested):
Good change, but needs some fixing in the linter.

mitch000001:
As the mechanism to verify if glance is up is using the import methods enabled does it make sense to also expose those as a metric?

niedbalski:
Please fix the conflicts before I can review. Thank you.

PR #254, opened 2022-09-21
Add deploy kernel and deploy ramdisk metrics to ironic exporter

dkt26111:
The deploy kernel and ramdisk are useful metrics when looking at an ironic node.
This PR adds the deploy kernel and deploy ramdisk as metrics for the ironic exporter.


niedbalski (approved):
LGTM, thanks a lot for your contribution.

PR #261, opened 2022-11-11
Fix neutron_port fixtures for gophercloud-1.0.0

martialblog:
Hi, this should fix the tests for gophercloud-1.0.0. 

I think it is caused by this change https://github.com/gophercloud/gophercloud/pull/2445 which adds the created_at/updated_at fields for the port resource but does not include logic to handle both types of timestamps.


martialblog:
Been testing the gophercloud-1.0.0 with a networking v2.0 API (Wallaby) and the neutron_ports metrics work as expected.

Issue #249, opened 2022-08-24
Lots of Nova metrics moved to Placement with Wallaby release (OS_COMPUTE_API >= 2.88)


frittentheke:
When targeting OpenStack Wallaby lots of Nova metrics ``openstack_nova_*``, like ``openstack_nova_memory_used_bytes`` are not returned anymore. This is due to the deprecation / move of quite a list of metrics for Nova towards Placement:

```
    current_workload
    cpu_info
    vcpus
    vcpus_used
    free_disk_gb
    local_gb
    local_gb_used
    disk_available_least
    free_ram_mb
    memory_mb
    memory_mb_used
    running_vms
```
see https://docs.openstack.org/nova/wallaby/reference/api-microversion-history.html#maximum-in-wallaby

Certainly pinning the API to the latest version still returning this data ( `OS_COMPUTE_API=2.87`) works,
but this adds to similar issues when using the exporter with more recent OS versions:

- https://github.com/openstack-exporter/openstack-exporter/issues/220
- https://github.com/openstack-exporter/openstack-exporter/issues/221

niedbalski:
@frittentheke are you already working on a fix for this?

frittentheke:
@niedbalski actually no, sorry.

StopMotionCuber:
Our division would benefit from getting this fixed, therefore I would take a look on this issue and try to get a PR ready

StopMotionCuber:
What is the preferred way to solve this issue? I'm seeing that most of the metrics _are_ exported, but at a different place within the placement API.

Possible Strategies:
 - Remove the old, non-functioning metrics in favor of the new ones.
 - For backwards-compatibility export the old metrics additionally under the `openstack_nova_` namespace.
   - Do this within `nova.go`, as metrics were previously published here
   - Do this within `placement.go`, as metrics are already queried here, so no additional API calls would be necessary

frittentheke:
@StopMotionCuber thanks for picking up on this issue !

To me this issue is mostly about adjusting the exporter to the API changes introduced with more recent OpenStack releases and I personally strongly favor option 1.

I know there are quite a few old OpenStack installations running out there and others might disagree with me. But I'd rather keep things clean and change the exporter to work with current OpenStack releases and break / drop compatibility for older releases in the process.

If you consult https://releases.openstack.org/#release-series everything older than Yoga is in EM and even Yoga is approaching extended maintenance soon. Additionally even EM is about to change it's guarantees and promises - see
[...]

PR #293, opened 2023-08-21
Move stats that are available via Placement API over

StopMotionCuber:
This is a first attempt in order to make some stats from #249 available again. Not really intended to get merged in this state, but to trigger some discussion

Unfortunately, it's not backwards compatible right now, as the name is changed e.g. from `openstack_nova_vcpus_available` to `openstack_placement_vcpus_available` due to the current implementation. But it also wasn't possible to query the information within `nova.go`, as the `exporter.Client` made some assumptions about the microversion, leading to a 404 when doing the queries within that file. I could investigate whether it is possible to give it another prefix than the intended one.

Additionally I'm not sure how the usage is measured for RAM and storage. Is that the actual RAM usage or the allocated RAM usage by VMs? I guess these differ to vCPUs, as they can be overallocated, but not overused


niedbalski:
@StopMotionCuber I am okay with reviewing this one, as long as you can check the test failures.

StopMotionCuber:
Done so. Do you have an idea how the fixtures were created? These seem outdated to me, as they still contain removed fields from the hypervisor API. I'd prefer to not edit them by hand and get the response somehow from a real world example with a recent version of OpenStack.

Updating these would lead to some test failures where metrics are missing. An even better solution would be to not use fixtures for these but actually fetch this information from a live OpenStack instance, but I haven't set one up myself yet and I don't know what kind of beast it is. Neither did I check how many of the metrics are flaky, something like RAM usage would be flaky for sure

frittentheke:
@niedbalski Regarding the breaking changes of "recent" OpenStack releases this is one of the changes that is required to export certain metrics again.

See my comment at https://github.com/openstack-exporter/openstack-exporter/issues/299#issuecomment-1895524583
I gladly test this if that helps you any in reviewing and merging this.

StopMotionCuber:
As taking a deeper look, my recommendation would actually not be to merge this one (it's kinda hacky) but rather merge https://github.com/openstack-exporter/grafana-dashboards/pull/1. Basically all required metrics are already there, they just have a new name.
You may already use that locally to have a working grafana.

frittentheke:
> As taking a deeper look, my recommendation would actually not be to merge this one (it's kinda hacky) but rather merge [openstack-exporter/grafana-dashboards#1](https://github.com/openstack-exporter/grafana-dashboards/pull/1). Basically all required metrics are already there, they just have a new name. You may already use that locally to have a working grafana.

If all those existing placement metrics already cover all of the missing ones formally fetched from nova (https://github.com/openstack-exporter/openstack-exporter/issues/249#issue-1349067683) that would really slick.

The only thing left would then be to remove the attempts to fetch the metrics from Nova in a recent version of the OpenStack exporter?

Issue #256, opened 2022-10-10
Log library prometheus/common/log is deprecated


Jonher937:
I have pinned prometheus/common to v0.26.0 in the meanwhile.
Deprecation note can be found [here](https://pkg.go.dev/github.com/prometheus/common@v0.26.0#section-readme)

PR #294, opened 2023-08-22
Update logging and dependencies

Knalltuete5000:
This PR includes multiple changes to the exporter, but should not change the functionality of the exporter.

First of all this PR upgrades the dependencies and removes the deprecated log library form prometheus mentioned in #256. It will be replaced with the github.com/go-kit/log library, which is used by other exporters, e.g. the [smartctl_exporter](https://github.com/prometheus-community/smartctl_exporter) or the [node_exporter](https://github.com/prometheus/node_exporter)

Due to the switch of the logging library it is possible to upgrade the prometheus libraries to the current versions. This includes also scraping the exporter via tls.

In addition the landing page is replaced with the landing page provided by the github.com/prometheus/exporter-toolkit

**NOTE**: I have added a golangci configuration file to exclude the `.Log`-method from checking the error returned. Please let me know if this is wanted or not. If not I will remove the configuration fill and work around the linting issues.


exporters/exporter.go
```diff
@@ -176,7 +178,7 @@ func (exporter *BaseOpenStackExporter) AddMetric(name string, fn ListFunc, label
 	// @TODO: get the region. constLabels["region"] = exporter.
 
 	if _, ok := exporter.Metrics[name]; !ok {
-		log.Infof("Adding metric: %s to exporter: %s", name, exporter.Name)
+		// log.Infof("Adding metric: %s to exporter: %s", name, exporter.Name)
```
niedbalski, line 181: please don't remove this.

main.go
```diff
@@ -46,68 +49,76 @@ func main() {
...
+		}
+		http.Handle("/", landingPage)
 	}
 
 	if *domainID != "" {
-		log.Infoln("Gathering metrics for domain ID", *domainID)
+		level.Info(logger).Log("msg", "Gathering metrics for configured domain ID", "domain_id", *domainID)
 	}
 
-	tcp := exporters.IP4or6(*bind)
```
niedbalski, line 100: why are you removing this?

niedbalski (changes requested):
Needs extra info here.

main.go
Knalltuete5000: The server as well as the config `web.listen-address` is handled by the [exporter-toolkit](github.com/prometheus/exporter-toolkit) [here](https://github.com/prometheus/exporter-toolkit/blob/e26aafa966f841a37eb9fcd91cc7ea8759e00225/web/tls_config.go#L300). It is using the default of the `IP4or6` method `tcp`. So there is no need for us to set the listen socket to a specified protocol. Hope it explains why it is removed.

PR #297, opened 2023-08-29
Fix flavor metrics

odnobit:
#### nova_flavor metrics fix
* nova_flavor metrics were collecting incorrectly because the `AccessType` parameter wasn't added.
openstack_exporter was only collecting flavors from the current project.
* Also, the fixtures/nova_os_flavors.json has been changed to correct one.


lobanov95:
@niedbalski greetings! 
please check, if you could.

odnobit:
@niedbalski hello! what do u think about changes?

niedbalski (approved):
LGTM

Issue #220, opened 2022-02-03
openstack exporter nova_server_status fails to return flavor id


jheikki100:
We are trying to get openstack metrics to Prometheus for further accounting data processing with cloudkitty.
However, when I query prometheus metrics from openstack exporter

`openstack_nova_server_status` field `flavor_id` is always returning \<nil> in prometheus openstack exporter.

For example:

```
openstack_nova_server_status{address_ipv4="",address_ipv6="",availability_zone="nova",flavor_id="<nil>",host_id="34473036769974153a388fb747e773fd7aa6095e48c6e23096b105c1",h
ypervisor_hostname="bm-compute01-lm",id="4177d13c-d42b-499c-a917-c76d2dc75226",name="demo-jhtest",status="ACTIVE",tenant_id="960168fa6ffd4b1d8d6007c3f234fa6c",user_id="d4fe
783e96a743fdba7d207b519802c7",uuid="4177d13c-d42b-499c-a917-c76d2dc75226"} 0
```
It seems that Prometheus is aware of flavors because direct query 
`curl 'http://prometheus_host:9091/api/v1/label/flavor_id/values`
returns list where values correspond to existing flavor ids in OpenStack.

Our OpenStack instance is currently on Wallaby release and is based on Kolla-Ansible.

Looking forward for suggestions or a fix because we really need the flavor_id in downstream accounting processing.

gtherond:
That's because the way the flavor are returned by the API changed with a microversion and that microversion was promoted as the minimum version for victoria.

Horizon had the same issue up to 18.6.4 that was recently patched: https://bugs.launchpad.net/horizon/+bug/1940834

priteau:
> Same problem with rocky. Output of v1.1.0 vs v1.5.0:
> 
> ```
> v1.1.0
> openstack_nova_server_status{address_ipv4="",address_ipv6="",availability_zone="ch-zh1-az2",flavor_id="xxxxx-4a78-8dc0-206deexxxxxxxx",host_id="c3165b602471b79356d6210a2775b0f6c9b268065292aff84a2615d8",id="xxxxxxxx",name="test-u1804-az2",status="ACTIVE",tenant_id="yyyyyyyyyyyyyyy",user_id="zzzzzzzzzzzz",uuid="idididididid"} 0
> 
> v1.5.0
[...]

PR #308, opened 2023-09-18
Fix <nil> values in nova_server_status

odnobit:
#220 server_status flavor_id label is always returning `nil` value:
    Starting from micro-version 2.47, `flavor.id` is no longer available in detailed server data.
    The data now only contains `flavor.name`.
    This fix involves retrieving all flavors and subsequently searching for flavor.id based on its name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1058,6 +1058,9 @@ openstack_nova_security_groups{region="Region"} 5.0
 # HELP openstack_nova_server_local_gb server_local_gb
 # TYPE openstack_nova_server_local_gb gauge
 openstack_nova_server_local_gb{id="27bb2854-b06a-48f5-ab4e-139817b8b8ff",name="openstack-monitoring-0",tenant_id="110f6313d2d346b4aa90eabe4970b62a"} 10
+# HELP openstack_nova_server_status server_status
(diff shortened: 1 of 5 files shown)


yo000:
Just to say, with patch applied, this bug can still occur if env var "OS_COMPUTE_API_VERSION" is set to non-numerical value like "latest"
Because of the test https://github.com/openstack-exporter/openstack-exporter/blob/44917926d4cc17effbf580cb8fcef4dcb3f7fe5b/exporters/nova.go#L313C1-L314C20

odnobit:
> Just to say, with patch applied, this bug can still occur if env var "OS_COMPUTE_API_VERSION" is set to non-numerical value like "latest" Because of the test https://github.com/openstack-exporter/openstack-exporter/blob/44917926d4cc17effbf580cb8fcef4dcb3f7fe5b/exporters/nova.go#L313C1-L314C20


sorry, but you can't use string value for `OS_COMPUTE_API_VERSION`.
example from openstack-cli
```bash
$ OS_COMPUTE_API_VERSION=latest openstack server list
id format of client version 'latest'. Expected format 'X.Y', where X is a major part and Y is a minor part of version.
```
Just use float value for that variable.

PR #321, opened 2024-02-02
Fix flavor_id="<nil>" when OS_COMPUTE_API_VERSION=latest

yo000:
Fix 'flavor_id="<nil>"' when "OS_COMPUTE_API_VERSION=latest" by getting actual API version like if no OS_COMPUTE_API_VERSION was set

### DIFF
--- a/exporters/nova.go
+++ b/exporters/nova.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


odnobit:
Please just don't use string value for `OS_COMPUTE_API_VERSION` variable.
It's not allowed. Thank you.

PR #327, opened 2024-02-22
Move channel to OFTC

Jonher937:
The openstack community moved over to OFTC about 3 years ago and the channel on Freenode is currently empty. On OFTC we are at least a few users that can answer questions.

https://opendev.org/openstack/openstack-manuals/commit/e168dfff03fc97b95d2197b9fe00d309d9c461f1


priteau:
I also hang out in this channel 👍

PR #329, opened 2024-02-23
Update VERSION to 1.7.0

frittentheke:


priteau:
I believe this needs to be bumped before release. Would we re-release 1.7.0?

niedbalski:
Be better to provide with a github action on tag that auto-updates this file as well?

frittentheke:
> Be better to provide with a github action on tag that auto-updates this file as well?

I actually did not even look what this file is used for / how it is used. I just spotted it not being of the latest version.
We did refactor the libvirt-exporter a while back and created some go-releaser + dependabot config via GitHub Actions
Maybe that could be of help / inspiration?

* https://github.com/inovex/prometheus-libvirt-exporter/blob/master/.goreleaser.yml
* https://github.com/inovex/prometheus-libvirt-exporter/tree/master/.github

^^ we also create the version string there, via the current tag.

Issue #351, opened 2024-03-15
[Feature Request] Cache mechanism


jneo8:
relate to: #214 

We propose introducing an optional caching feature to the exporter to address slow queries, which can be activated as follows:

```
# Cache metrics by name
openstack-exporter --cached-metrics=AAA,BBB

# Cache metrics by service 
openstack-exporter --cached-services=AAA,BBB

# Global cache
openstack-exporter --cached --cached-ttl 60s
```

This feature would entail a background process that collects metrics every 5 minutes (customizable) and stores them in a file or memory. Cached values would be returned upon Prometheus scraping, provided the cache is ready.

We're eager to assist with its implementation but seek the maintainer's approval on whether this would be a valuable addition.

> Note: Our main purpose is to have a cache mechanism, so not necessarily implemented as described above

fprzewozny:
That would be cool!

jneo8:
Hi Christian. Thanks for your feedback. I want to make sure we are on the same page here.

> But if the exporter was to collect all the data independently from the Prometheus scraping I suggest to also introduce some kind of worker pattern to split up the collection of all those metrics.

When you mention about `split up the collection of all those metrics`, what is the granularity here? We separate by the service and run query parallel? Or you want to have a more small granularity(which means re-write the collector)?

I think we can limit the scope here, we still use the same collector and provide a cache here. The collector optimization can be another issue. How do you think?

(We also thinking about the scaling problem, so it's nice if we can make it scalable)

PR #355, opened 2024-03-29
feat: Basic version of cache mechanism

jneo8:
Fix: #351 

- A background service is responsible for gathering metrics and caching them, incorporating a mechanism for Time-to-Live (TTL) expiration.
  - The collection task is scheduled to run at every half interval of the TTL.
- New cache package provides cache backend interface and in-memory store implementation
- metricHandler and probeHandler read caches then return if cache is enabled

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -9,7 +9,6 @@
 package cache
 
 import (
-	"fmt"
 	"sync"
 	"time"
 
@@ -24,36 +23,20 @@ type CacheBackend interface {
 	SetCloudCache(string, CloudCache)
 	// Get cache with cloud name.
 	GetCloudCache(string) (CloudCache, bool)
-	// Set cache with service name.
-	SetServiceCache(string, string, ServiceCache)
-	// Get cache with service name.
-	GetServiceCache(string, string) (ServiceCache, bool)
-	// Set cache with metric family name.
-	SetMetricFamilyCache(string, string, string, MetricFamilyCache)
-	// Get cache with metric family name.
-	GetMetricFamilyCache(string, string, string) (MetricFamilyCache, bool)
 	// Flush expired caches based on cloud's update time.
 	FlushExpiredCloudCaches(time.Duration)
-	// Flush all cache data
-	Clear()
 }
 
 // MetricFamily Cache Data
 type MetricFamilyCache struct {
-	Time time.Time
-	MF   *dto.MetricFamily
-}
-
-// Service Cache Data
-type ServiceCache struct {
-	Time               time.Time
-	MetricFamilyCaches map[string]*MetricFamilyCache
+	Service string
+	MF      *dto.MetricFamily
 }
 
 // Cloud Cache Data
 type CloudCache struct {
-	Time          time.Time
-	ServiceCaches map[string]*ServiceCache
+	Time               time.Time
+	MetricFamilyCaches map[string]*MetricFamilyCache
 }
 
 // GetCache return a singleton CacheBackend
@@ -75,7 +58,7 @@ type InMemoryCache struct {
 }
 
 // init ensures the values are not missing in the nested map.
-func (c *InMemoryCache) init(cloud *string, service *string, mfName *string) {
+func (c *InMemoryCache) init(cloud *string) {
 	if c.CloudCaches == nil {
 		c.CloudCaches = make(map[string]*CloudCache)
 	}
@@ -84,7 +67,6 @@ func (c *InMemoryCache) init(cloud *string, service *string, mfName *string) {
 			cloudCache := NewCloudCache()
 			c.CloudCaches[*cloud] = &cloudCache
 		}
-		c.CloudCaches[*cloud].Init(service, mfName)
 	}
 }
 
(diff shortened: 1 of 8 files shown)


cache/cache.go
```diff
@@ -0,0 +1,185 @@
...
+var singleCache CacheBackend
+var once sync.Once
+
+type CacheBackend interface {
+	// Set cache with cloud name.
+	SetCloudCache(string, CloudCache)
+	// Get cache with cloud name.
+	GetCloudCache(string) (CloudCache, bool)
+	// Set cache with service name.
+	SetServiceCache(string, string, ServiceCache)
```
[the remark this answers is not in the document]
jneo8: I would like to keep those methods. Let's see other's feedback.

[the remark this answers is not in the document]
niedbalski: @jneo8 If unused, please remove.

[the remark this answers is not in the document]
jneo8: They are used in the unit-tests.
Remove some of them will make the code very hard to test.

For example if I want to test `SetMetricFamilyCache`, I need to have a work-around way to get the cache object from maybe server or cloud level then verify it's exists, which also make testing code really dirty.

Also if you remove `GetCloudCache`, it will somehow remove the work-around way also.

Since those functions are just get/set methods and either been used in the code or test code. I would like to keep them in the interface.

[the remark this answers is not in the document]
jneo8: The refactor is done in the [newest commit](https://github.com/openstack-exporter/openstack-exporter/pull/355/commits/b89be8c183a77b71a39c5f08654daf59beda23b2)

cache/utils.go
```diff
@@ -0,0 +1,136 @@
...
+	enabledServices := []string{}
+	for service, disabled := range services {
+		if !*disabled {
+			enabledServices = append(enabledServices, service)
+		}
+	}
+
+	for _, cloud := range clouds {
+		level.Info(logger).Log("msg", "Start update cache data", "cloud", cloud)
+		// Update cloud's cache once finish all exporters' collection job.
```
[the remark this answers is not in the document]
jneo8: Updated

cache/utils.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/gophercloud/utils/openstack/clientconfig"
+	"github.com/openstack-exporter/openstack-exporter/exporters"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/expfmt"
+)
+
+// CollectCache collects the MetricsFamily for required clouds and services and stores in the cache.
+func CollectCache(
+	enableExporterFunc func(string, string, string, []string, string, bool, bool, bool, bool, string, func() (string, error), log.Logger) (*exporters.OpenStackExporter, error), multiCloud bool, services map[string]*bool, prefix, cloud string, disabledMetrics []string, endpointType string, collectTime bool, disableSlowMetrics bool, disableDeprecatedMetrics bool, disableCinderAgentUUID bool, domainID string, uuidGenFunc func() (string, error), logger log.Logger) error {
```
chanchiwai-ray, line 22: Is there no other way to write this? I'm having trouble comprehending it

cache/utils.go
```diff
@@ -0,0 +1,152 @@
...
+
+	for _, cloud := range clouds {
+		level.Info(logger).Log("msg", "Start update cache data", "cloud", cloud)
+		// Update cloud's cache once finish all exporters' collection job. so we won't mix the old
+		// and new metrics in the cache and confuse users.
+		cloudCache := NewCloudCache()
+
+		for _, service := range enabledServices {
+			level.Info(logger).Log("msg", "Start collect cache data", "cloud", cloud, "service", service)
+			exp, err := enableExporterFunc(service, prefix, cloud, disabledMetrics, endpointType, collectTime, disableSlowMetrics, disableDeprecatedMetrics, disableCinderAgentUUID, domainID, nil, logger)
```
chanchiwai-ray, line 71: Since you imported the `exporters`, can you just use `exporters.EnableExporter` instead of the anonymous `enableExporterFunc` argument ?

cache/utils.go
```diff
@@ -0,0 +1,152 @@
...
+			if err != nil {
+				// Log error and continue with enabling other exporters
+				level.Error(logger).Log("err", "enabling exporter for service failed", "cloud", cloud, "service", service, "error", err)
+				continue
+			}
+			registry := prometheus.NewPedanticRegistry()
```
chanchiwai-ray, line 77: iiuc, this should be outside the inner for-loop, otherwise you create a registry for every service, and that `Gather` method only gather one service for each service in each cloud

cache/utils.go
jneo8: It's a practice to inject the func so you can more easy to write the unit test.

See https://github.com/jneo8/openstack-exporter/blob/fb699a36e0eba85590b5a311d467259766d9a8a3/cache/utils_test.go#L88

cache/utils.go
jneo8: You can't do that here, let me explain why:

Imaging you put `prometheus.NewPedanticRegistry` in first for loop

```
for _, cloud := range clouds {
        registry := prometheus.NewPedanticRegistry()
        ...
        for _, service := range enabledServices {
                // If you have multiple services, the new exporter will append to the registry's exporter list and there is no way the flush it.
                // So the same metrics will run collect multiple times during the for loop iteration.
                registry.mustRegister(*exp)
                metricFamilies, err := registry.Gather()
                ...
        }
}
```

chanchiwai-ray: The `Gather` should be called outside the inner `for-loop` though, i.e. when you finish registering all the exporter in the same cloud, at least that's how python prometheus client works, not sure if it's the same in Go

cache/utils.go
chanchiwai-ray: Okay, understood. but not sure about that: should we make the actual code better or unit test better. (not an expert in Go). Feel free to resolve because I can't

cache/utils.go
chanchiwai-ray: OIkay thanks for the clarification, it seems a design decision to cache per service. Please consider this resolved

jneo8:
Hi @niedbalski

Can we get your review again?

Pjack:
Hi @niedbalski , may I get your review ? If there is any concern, please let us know. thanks!

niedbalski (approved):
Looks a good contribution. +1

PR #380, opened 2024-07-22
Add TenantID argument and use it in nova and cinder exporters

Krast76:
Hi there,

First of all, thanks for this exporter it is really useful.

As an OpenStack's user (not an admin one), I want to use the exporter on a scoped tenant.

To do so, I've added tenantID argument and use it on the nova and cinder exporters to list all the instances and volumes for a given tenant.

Regards,

### DIFF
--- a/cache/utils.go
+++ b/cache/utils.go
@@ -20,7 +20,7 @@ import (
 // CollectCache collects the MetricsFamily for required clouds and services and stores in the cache.
 func CollectCache(
 	enableExporterFunc func(
-		string, string, string, []string, string, bool, bool, bool, bool, string, func() (string, error), log.Logger,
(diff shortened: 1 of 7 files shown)


Sharpz7 (approved):
LGTM

It looks like you would have tested this in a devstack/prod server given your background, so I think the unit tests passing is enough.

Krast76:
@Sharpz7 Exactly ! I run the code from this PR since almost three weeks now.

I have rebase my branch from main and all unit tests seems ok.

Sharpz7 (approved):

Krast76:
Is there something else I must do to get this merged ?

niedbalski (approved):

PR #393, opened 2024-07-31
Update deps

Jonher937:
Updated all deps excluding github.com/prometheus addressed in #388

This change is to prepare for additional work on the exporter being done against updated versions of llibraries.


Sharpz7 (approved):
Figured this out seperately, was happy to see that someone else had already fixed it.

Thank you :)

PR #398, opened 2024-08-20
Jocke fix hypervisor

jhindersson:


Jonher937:
Was this meant for a internal fork? This does not look like changes that we'd want upstream.

Issue #206, opened 2021-10-12
Running with restricted privileges


mkszuba:
Hello,

If I understand your documentation correctly, openstack-exporter is supposed to connect to OpenStack as a user with full admin rights in the projects under consideration. Sadly, this is at odds with security policies of our sites. Is there a list somewhere of what the exact privileges the exporter user must have?

gtherond:
> Hello,
> 
> If I understand your documentation correctly, openstack-exporter is supposed to connect to OpenStack as a user with full admin rights in the projects under consideration. Sadly, this is at odds with security policies of our sites. Is there a list somewhere of what the exact privileges the exporter user must have?

You could indeed use a reader system scoped role and so at least limit security issues in terms of behavior if not in terms of discoverability.

tabacha:
Hi, we have created an openstack with default roles.

When we use admin user, everything works.
We created a user monitor:
```
openstack user create --password mypassword monitor
+---------------------+----------------------------------+
| Field               | Value                            |
+---------------------+----------------------------------+
| domain_id           | default                          |
| enabled             | True                             |
| id                  | 4336f553239140949b0e0226b4ec19e0 |
| name                | monitor                          |
| options             | {}                               |
| password_expires_at | None                             |
+---------------------+----------------------------------+
openstack role add --system all --user monitor reader
openstack role assignment list --names --system all
+---------------------+-----------------+--------------------------+---------+--------+--------+-----------+
| Role                | User            | Group                    | Project | Domain | System | Inherited |
+---------------------+-----------------+--------------------------+---------+--------+--------+-----------+
| admin               |                 | OpenStack Admins@Default |         |        | all    | False     |
| load-balancer_admin |                 | OpenStack Admins@Default |         |        | all    | False     |
| admin               | admin@Default   |                          |         |        | all    | False     |
| reader              | monitor@Default |                          |         |        | all    | False     |
+---------------------+-----------------+--------------------------+---------+--------+--------+-----------+
```

When we stated the exporter:
```
Jul 21 09:40:57 monitor01 golang-openstack-exporter.openstack-exporter[1112787]: time="2022-07-21T09:40:57Z" level=info msg="Build context (go=go1.16.3, user=, date=)" source="main.go:61"
[...]

PR #424, opened 2025-01-13
feat: consistent tenantID filter and listProjects only on no tenantID specified

stijzermans:
To have the exporter work in situations where only permissions for a single project are given, `list_projects` permissions are often not handed out.

Hence created a helper GetProjects function which will not perform ListProjects a tenantID specified.

This should help along with #206 for situations where permission are only project scoped.


stijzermans:
Thanks @niedbalski for moving it to ready :) did some testing today and made some final adjustments.
This is now ready for review 😄 .

stijzermans:
@niedbalski I'll gladly help out if wanted with updating the GitHub actions, but for I think the pipeline is no longer blocking on the current changes. Anything I can do to help moving forward with this change?

Sharpz7 (approved):

PR #433, opened 2025-03-07
Paginate ListAllMetrics function

omid-ak:
Implement pagination in the ListAllMetrics function to efficiently count metrics without loading all of them into memory. This change improves performance and reduces memory usage.

### DIFF
--- a/exporters/gnocchi.go
+++ b/exporters/gnocchi.go
@@ -2,6 +2,7 @@ package exporters
 
 import (
 	"github.com/go-kit/log"
+	"github.com/gophercloud/gophercloud/pagination"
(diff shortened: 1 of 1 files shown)


Sharpz7:
I am not convinced this is needed, do metrics really take that much space in memory to justify this change? I would imagine most of the time metrics are being consumed immediately into another storage system.

omid-ak:
You're right — the metrics are indeed being ingested directly into Gnocchi storage. I'm referring specifically to clusters with over 5,000 instances, where the exporter can run into OOM (Out of Memory) issues under this load. I personally encountered this problem and resolved it with a fix, which is why I decided to contribute it back to the original codebase.

Issue #430, opened 2025-03-02
Concurrent map writes and flawed ListComputeLimits logic cause exporter crashes and incorrect Identity client usage


Tatikoma:
**Title**: Concurrent `map` writes and flawed `ListComputeLimits` logic cause exporter crashes and incorrect Identity client usage

---

### Summary

When adding logic to expose Nova limits (for quota monitoring and capacity planning), the code introduced two issues:

1. **Concurrent writes to a global `map`** (`endpointOpts`) cause a Go runtime panic (`concurrent map writes`) under load.
2. **Flawed usage** of `endpointOpts` in `ListComputeLimits` can produce **incorrect** Identity client configuration because each new request may overwrite data in the global map.

---

#### 1. Concurrent Map Writes

- The exporter uses a **global map** `endpointOpts` to store `gophercloud.EndpointOpts` for various services.
- Functions like `NewServiceClient` write to this map (`endpointOpts[service] = eo`) every time a client is created.
- Under concurrent scrapes, multiple goroutines can write to `endpointOpts` simultaneously, triggering a Go runtime panic: `fatal error: concurrent map writes`
- Go maps are not thread-safe, and **any** unsynchronized parallel writes will eventually crash the exporter.

#### 2. Flawed Logic in `ListComputeLimits`

- The `ListComputeLimits` function attempts to create a Keystone (`identity`) client by checking:
```go
if v, ok := endpointOpts["identity"]; ok {
    eo = v
} else if v, ok := endpointOpts["compute"]; ok {
    eo = v
} else {
    return errors.New("No EndpointOpts available to create Identity client")
}
```

Because `endpointOpts` is global (and overwritten on each new client creation), **the stored data might not match** the actual identity or region settings intended for the current request.
This logic can produce incorrect or mismatched endpoint configurations if different scrapes or different services overwrite the same entry (`endpointOpts["identity"]` or `endpointOpts["compute"]`).
Even with serialization (a mutex), different concurrent requests can overwrite `endpointOpts["identity"]` in turn. As a result, `ListComputeLimits` may use the wrong or outdated settings, leading to errors or inconsistent data.

---

### Impact

- **Crash**: Concurrent requests eventually crash the entire exporter.
- **Incorrect Identity Config**: Even without a crash, the map-based approach means a region or availability setting can be silently replaced by another request, causing `ListComputeLimits` to query the wrong endpoint (or fail altogether).

[...]

PR #434, opened 2025-03-10
fix concurrent map writes crash (#1)

Tatikoma:
Fix crash related to https://github.com/openstack-exporter/openstack-exporter/issues/430

I have tested this fix in production and it solved crash issue.


alexeymyltsev:
LGTM

PR #455, opened 2025-06-20
Adding Tenant ID to shares

ImSundee:
Tenant ID is currently missing from Share info, adding this just to enable showing shares by Tenant when required.


Sharpz7:
I think this can be closed with #457

PR #462, opened 2025-06-27
Fixed Goreleaser

Sharpz7:
It seems that there are breaking changes

https://github.com/goreleaser/goreleaser-action?tab=readme-ov-file#migrating-from-v3 https://github.com/orgs/goreleaser/discussions/2786


Sharpz7:
I will note I have not tested this locally, but I think it is valid?

Sharpz7:
@mnaser we need to get this one merged to fix CI.

PR #466, opened 2025-06-27
Use cpu_info instead of vcpus to get the number of vcpus

drencrom:
The total number of VCPUs should be obtained from cpu_info instead of the vcpus field. This vcpus value changes if there are pinned, dedicated or shared CPUs configured in nova and, crucially, can be less than vcpus_used.

The cpu_info data may be empty in some openstack versions, so fallback to old method in that case.


drencrom:
Hi folks, can any of you take a look at this patch?

niedbalski (approved):

drencrom:
Hi, I changed the patch for a signed one as that is required for this project.

drencrom:
Hi @niedbalski Could you please review this commit again as I re-commited it with a signature?
Thanks!

Sharpz7 (dismissed):

drencrom:
Hi @Sharpz7 Please check if it ok now.

nicolasbock (approved):
LGTM

PR #392, opened 2024-07-31
Add Nova metadatas into labels for server status metric

sonvuthai:
**What this PR does:**
Inspiring by #376 and Prometheus OpenStack Service Discovery, this will append metadata per instance into server_status metric label.
With this change, i think it could be a better way for querying metric than convert all the metadata into one label value.

PR #493, opened 2025-10-08
Add nova metadata to labels mapping

vooon:
It's alternative to #392 - but carry only preconfigured set of labels, which helps to reduce label index cardinality.

Basically that adds `--nova.metadata-extra-labels` option, which controls what metadata keys should be translated into a label.
Format is `label_name=metadata_key`, or just `metadata_key` when both sides are equal.
Option is repeatable, so you can provide multiple mappings that way, or just use `,` to combine a list.

### DIFF
--- a/cache/utils.go
+++ b/cache/utils.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/gophercloud/utils/openstack/clientconfig"
 	"github.com/openstack-exporter/openstack-exporter/exporters"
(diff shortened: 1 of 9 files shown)


Sharpz7:
@vooon I apologies, Github Copilot is automated commenting. I am not sure why it is getting triggered. I have dismissed its comments as they are not really useful.

Sharpz7 (approved):

PR #419, opened 2024-11-06
Add `ListNetworkLimits` func

JustHumanz:
resolved #418
```bash
╭─[humanz-403] as humanz in ~                                                                                                                          15:24:34
╰──➤ curl -s http://localhost:9180/metrics | grep neutron | grep limits
# HELP openstack_neutron_limits_floating_ip_max limits_floating_ip_max
# TYPE openstack_neutron_limits_floating_ip_max gauge
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="174baba8955e43aca85121f80fe2b57a"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="f79a30cc3b0841229bbe7e43bb56611b"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="c99ef2be8a6548afb48ee2ea2aed46cc"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="53871ce866994d11ad607327463a599f"} 1
openstack_neutron_limits_floating_ip_max{tenant="XXXXXXXX",tenant_id="1345ae46d69b4d83ad330cffb4ec22bb"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXXXXXXXXXX",tenant_id="79abc434355841b6b0351dca122a2660"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="de8ef4d60e5c4e2583236eee95287936"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="65e8fc1b682a40cb8839fc1beb24927d"} 1
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="fb8a24128f594cc99eca4b0bc369536c"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXX",tenant_id="1afb6a7fbcc04f2caea9f3fddeea4fd0"} 5
openstack_neutron_limits_floating_ip_max{tenant="XXXXXXXX",tenant_id="9d418b38afa14cc8a13ac6b8131975d0"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXXn",tenant_id="4ad90163ef2c4ae09c16a523dc40a234"} 50
openstack_neutron_limits_floating_ip_max{tenant="XXXXyanoble",tenant_id="997f4904c1fd4ff585e9c28af4aaeef8"} 50
[...]

PR #497, opened 2025-10-16
Add network limits and align code structure with PR #390

EnDjeee:
This PR expands on PR #419 by aligning the code with the structure introduced in PR #390.

This also implements tests for the network limits to cover the new code.


Sharpz7:
Hey @EnDjeee thanks. Honestly, I would rather get this merged than #419 if you agree that is suitable?

EnDjeee:
> Hey @EnDjeee thanks. Honestly, I would rather get this merged than #419 if you agree that is suitable?

Yes, i agree. The only remaining blocker for merging seems to be a non‑verified commit from JustHumanz. How would you like to proceed?

EnDjeee:
> Great! Thanks, does that mean #419 can be closed when this PR merges? (I am pretty sure the answer is yes but just want to confirm)

Yes, i believe #419 can be closed now.

PR #514, opened 2025-11-30
Fix Snapcraft publishing

Sharpz7:
Part of #600.

Build and publish the release Snap to the Snap Store when a version tag is pushed.


Sharpz7:
@niedbalski just following up again. Are we able to get access to the snapcraft API key to start using this?

Sharpz7:
@niedbalski just checking in again. I think if it takes much longer I will consider a new account.

PR #505, opened 2025-10-31
Port to gophercloud v2

vooon:
That is a refactoring to use `gophercloud/v2`.

- Update Go requirement to 1.25
- Drop gophercloud v1
- Fix #494 
- Added API discovery reports fixture to most of the services
- Add test for Swift
- Add build_info metric

Tries to solve some of #500

PR #520, opened 2025-12-15
fix: Avoid paging in ListHypervisors

### LINKED ISSUE #519: Hypervisors list paging is broken
Hello,

Hypervisors list is broken for quite a while for lists with more than the default max 1000 items (somewhere after v1.1.0). It is related to the introduction of paging in gophercloud. The exporter sets the micro version to the max supported one reported by the API, which include paging, but gophercloud v1 doesn't support paging. One way to circumvent this is to explicitly set a compute microversion through the dedicated environment variable to one not supporting paging (microVersion < 2.33, eg: OS_COMPUTE_API_VERSION=2.32). The issue with this is that it's sets this lower version for all compute API calls which might cause other problems. 

After toying with this, it seems to me the cleanest and easiest way to fix this is to use gophercloud v2, which already contains the paging fix, for the Hypervisors.List function.

Regards.

dsanader:
Proposal to fix #519

### DIFF
--- a/exporters/nova.go
+++ b/exporters/nova.go
@@ -180,10 +180,14 @@ func ListHypervisors(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metri
 	var allHypervisors []hypervisors.Hypervisor
 	var allAggregates []aggregates.Aggregate
 
+	// Explicitely set client to a version not supporting paging since paging is kinda broken
+	mv := exporter.Client.Microversion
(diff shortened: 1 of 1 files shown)


dsanader:
At first glance #505 includes the same fix. If #505 is merged, you should be able to safely drop this PR. I will build the #505 code just to check though and report back If I find anything wrong on my side.

Sharpz7:
Appreciate the detailed breakdown @dsanader . Thank you!

PR #524, opened 2026-01-13
Fallback to user scoped project listing when domain project listing is unauthorized

melonlorrd:
This change adds a fallback mechanism for project listing in the OpenStack exporter, if the user used lacks permission to list all domain projects (403 Unauthorized), the exporter will instead list projects scoped to the user.

I have made this change after analyzing the fallback flow of openstack-cli, please let me know if any changes are needed.


Sharpz7 (dismissed):
I am not 100% about this being treated as a fallback, but I also see the benefit for backwards compatibility.

Sharpz7:
Hey @melonlorrd, would you be able to fix these conflicts?

melonlorrd:
Hi @Sharpz7, I’ve resolved the merge conflicts and moved the user scoped project listing section to utils.go. Please let me know if any further changes are needed from my side. Thanks.

Sharpz7 (approved):

n2N8Z:
The Openstack v6.6.0 CLI has such a fallback, but somewhere between that and v9.0.0 that fallback was dropped.
Both versions however have the --my-projects switch that lists the projects the user has access to.
So for consistency with the CLI, rather than having an automatic fallback, perhaps an explicit switch requesting the user projects ?

PR #545, opened 2026-04-06
Openstack Designate Speedup

Sharpz7:


Sharpz7:
This PR is waiting on https://github.com/gophercloud/gophercloud/pull/3688

Sharpz7:
Keeping as draft because we need to get the v2 in and merged

PR #481, opened 2025-09-03
Added integration tests

Sharpz7:
This should help make trusting new changes easier.

PR #546, opened 2026-04-06
Placement Parallel Speedup

Sharpz7:
This pull request introduces a new feature to enable parallel collection of OpenStack Placement resource provider data, controlled by a new command-line flag. It adds support for running Placement API calls in parallel, which can significantly improve the performance of metric collection, especially in environments with many resource providers. The change is backward-compatible and can be toggled with the `--complete-placement-in-parallel` flag.

This is after observing this can take many minutes with many servers. In the long run, we probably want a change to placement directly to add a bulk API


Sharpz7:
This needs to be tested in a devstack / as part of the [integration tests](https://github.com/openstack-exporter/openstack-exporter/pull/481)

exporters/placement.go
```diff
@@ -80,9 +99,93 @@ func ListPlacementResourceProviders(exporter *BaseOpenStackExporter, ch chan<- p
...
+				prometheus.GaugeValue, float64(v.AllocationRatio), data.Name, k)
+
+			ch <- prometheus.MustNewConstMetric(exporter.Metrics["resource_reserved"].Metric,
+				prometheus.GaugeValue, float64(v.Reserved), data.Name, k)
+		}
+
+		for k, v := range data.Usages {
+			ch <- prometheus.MustNewConstMetric(exporter.Metrics["resource_usage"].Metric,
+				prometheus.GaugeValue, float64(v), data.Name, k)
+		}
```
vooon, line 187: I do not see any cross-talk on those two metrics, so probably no need to carry the data over channels from retrieval goroutine to this consumer - you can push metric directly to ch.

main.go
```diff
@@ -126,7 +127,7 @@ func cacheBackgroundService(ctx context.Context, services map[string]*bool, errC
 	defer ttlTicker.Stop()
 
 	// Collect cache data in the beginning.
-	if err := cache.CollectCache(exporters.EnableExporter, *multiCloud, services, *prefix, *cloud, *disabledMetrics, *endpointType, *collectTime, *disableSlowMetrics, *disableDeprecatedMetrics, *disableCinderAgentUUID, *domainID, *tenantID, novaMetadataMapping, nil, logger); err != nil {
+	if err := cache.CollectCache(exporters.EnableExporter, *multiCloud, services, *prefix, *cloud, *disabledMetrics, *endpointType, *collectTime, *disableSlowMetrics, *disableDeprecatedMetrics, *disableCinderAgentUUID, *domainID, *tenantID, novaMetadataMapping, nil, *completePlacementInParallel, logger); err != nil {
```
vooon, line 130: Options list became uglier each time... Looks like should be converted into struct.

exporters/placement.go
```diff
@@ -49,18 +62,24 @@ func ListPlacementResourceProviders(exporter *BaseOpenStackExporter, ch chan<- p
...
+		return nil
+	}
+
+	if exporter.CompletePlacementInParallel {
+		return listPlacementResourceProvidersParallel(exporter, ch, allResourceProviders)
+	}
+	return listPlacementResourceProvidersSequential(exporter, ch, allResourceProviders)
+}
 
+func listPlacementResourceProvidersSequential(exporter *BaseOpenStackExporter, ch chan<- prometheus.Metric, allResourceProviders []resourceproviders.ResourceProvider) error {
```
vooon, line 75: I think no need to keep two code paths. This sequential read is just a parallel with concurrency=1 :)

Sharpz7:
```
Benchmark	Without #546	With #546	Result
Cold collect, parallel enabled	2322.1 ms	465.7 ms	~5.0x faster
Warm collect, parallel enabled	2352.6 ms	215.5 ms	~10.9x faster
Direct list, parallel enabled	2136.9 ms	621.6 ms	~3.4x faster
Exporter cache write	94.0 ms	133.5 ms	roughly similar/slower
```

On top of https://github.com/openstack-exporter/openstack-exporter/pull/580

vooon (commented):
- [P1] Cached usage and allocation metrics become stale indefinitely. The cache considers a provider valid solely when cached.generation == rp.Generation (placement.go#L936-L944), but it caches resource_usage and resource_provider_allocations too (placement.go#L1110-L1154). Placement tracks allocation changes using consumer generations separately from resource-provider generations; allocation updates can therefore occur without changing rp.Generation. ([Placement API docs](https://docs.openstack.org/placement/2026.1/doc-placement.pdf?utm_source=chatgpt.com)) VM lifecycle changes could leave Prometheus reporting old usage/allocation data until an inventory/provider update occurs.

Consider caching only relatively static inventory/trait data, or refreshing usage and allocations on every scrape with an appropriate TTL/invalidation mechanism.

- [P1] The global cache can mix clouds sharing a Placement endpoint. Cache identity is only endpoint + collectTraits (placement.go#L805-L829). Two configured clouds using the same URL but different credentials, regions, or policy scopes will share provider data. If UUID and generation match, the second cloud can emit metrics populated by the first cloud’s cache.

Key the cache by the cloud/auth scope as well, or make it instance-scoped rather than global.

- [P2] Trait API failures are cached as successful empty traits. GetTraits errors are logged and ignored (placement.go#L1057-L1082); the provider is then emitted and cached with an empty trait label (placement.go#L1130-L1133). Because the cache is generation-based, a transient failure can persist until the provider generation changes.

Treat trait retrieval failure as a provider collection failure, or avoid caching that provider when traits were not successfully retrieved.

Assisted-by: Codex

PR #550, opened 2026-04-28
snap: add tenant id to the snap configuration hook

sergiocazzolato:
After installing and trying the snap, I see it needs to allow the configuration for the tenant id as it is required in some openstack environments.


Sharpz7:
Hey @sergiocazzolato, out of interest, which snap have you been trying to install?

sergiocazzolato:
Hi @Sharpz7, I built the golang-openstack-exporter snap, and it can't connect to our openstack deployment if we don't specify a tenant, iirc it couldn't retrieve info for all the tenants without admin permisions.

Sharpz7:
I think the last update to that snap was 2021, see https://github.com/openstack-exporter/openstack-exporter/pull/514


We are looking to get snap back to up-to-date ASAP

Sharpz7 (approved):

Issue #548, opened 2026-04-08
Integration Test Improvements


Sharpz7:
- Coverage for checking metric changes on gophercloud client changes (i.e. add a value to a node, check openstack exporter catches the change)
- Comments from @vooon in https://github.com/openstack-exporter/openstack-exporter/pull/481

PR #551, opened 2026-05-06
Ironic Fixes in CI / moving more in line with Gophercloud

Sharpz7:
Based on https://github.com/gophercloud/gophercloud/blob/main/.github/workflows/functional-baremetal.yaml#L65

A fix with the azure mirror and the `qemu-system-misc` package, plus changes to put us more in line with the YAML from gophercloud. This will be done in the other services as part of https://github.com/openstack-exporter/openstack-exporter/issues/548

### DIFF
--- a/.github/workflows/integration-baremetal.yml
+++ b/.github/workflows/integration-baremetal.yml
@@ -21,10 +21,25 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Configure apt network retries
+        run: |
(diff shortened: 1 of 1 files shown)


jayofdoom:
I troubleshooted with @Sharpz7 -- this looks like a temporary azure mirror issue, so workaround makes sense.

jayofdoom (approved):
Just noting this has additional changes, including stuff like enabling openstack-cli-server. It all looks good to me.

PR #557, opened 2026-05-24
Add neutron VPNaaS support

flashvoid:
New metrics are

* vpn_endpoint_groups
* vpn_ike_policies
* vpn_ipsec_policies
* vpn_services
* vpn_service
* vpn_siteconnections
* vpn_siteconnection

PR #559, opened 2026-05-29
Networking integration tests & applying DRY

Sharpz7:
Refactors integration test client creation to reduce duplication and simplify maintenance.

- Added shared helpers in `integration/clients/clients.go` for authenticated and unauthenticated service client creation.
- Updated existing service client constructors to use the new helpers and standardize region handling.
- Simplified `integration/baremetal_test.go` by using shared test helpers for exporter startup, metric scraping, and assertions.
- Updated the networking integration workflow to enable VPNaaS support and use `strongswan`.
- Removed unused imports and minor formatting issues.


Sharpz7:
Let's not merge just yet, and add tests for https://github.com/openstack-exporter/openstack-exporter/pull/557 after it merges, since it is neutron/networking as well.

I don't think the tests belong in the existing code because there are a lot of code changes needed to make it work

Sharpz7:
Hey @vooon I layed out a wider vision in https://github.com/openstack-exporter/openstack-exporter/discussions/585.

This is the PR that is blocking a lot of that progress. Are you able to give it a quick review?

integration/funcs/networking.go
```diff
@@ -0,0 +1,653 @@
...
+// RequireVPNaaSExtension skips the test when the Neutron VPNaaS extension is
+// not available.
+func RequireVPNaaSExtension(t *testing.T, client *gophercloud.ServiceClient) {
+	t.Helper()
+
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	if _, err := extensions.Get(ctx, client, "vpnaas").Extract(); err != nil {
+		t.Skipf("Neutron VPNaaS extension is not available; enable neutron-vpnaas in DevStack: %v", err)
```
vooon: Probably should only skip if vpnaas really disabled, and not some random 500?

integration/networking_test.go
```diff
@@ -1,91 +1,260 @@
...
+		metrics.requireUp(t, "openstack_neutron_up")
 	})
 
 	t.Run("openstack_neutron_core_metrics_present", func(t *testing.T) {
-		expected := []string{
+		metrics.requireAnyFamily(t,
 			"openstack_neutron_networks",
 			"openstack_neutron_ports",
 			"openstack_neutron_subnets",
 			"openstack_neutron_router",
```
vooon, line 35: I think here you should use something like `requireAllFamily()`, as all of that things must present?

integration/funcs/networking.go
```diff
@@ -0,0 +1,653 @@
...
+
+	networkName := tools.RandomString("ACPTTEST", 16)
+	createOpts := networks.CreateOpts{
+		Name:         networkName,
+		AdminStateUp: gophercloud.Enabled,
+	}
+
+	t.Logf("Attempting to create network: %s", networkName)
+
+	network, err := networks.Create(context.TODO(), client, createOpts).Extract()
```
vooon: There are lot's of `context.TODO()`, maybe better to introduce some helper to make context with timeout inherited from [`t.Context()`](https://pkg.go.dev/testing#T.Context)?

integration/funcs/networking.go
```diff
@@ -0,0 +1,653 @@
...
+	t.Logf("Deleted VPN service: %s", serviceID)
+}
+
+// CreateVPNSiteConnection creates a VPNaaS IPsec site connection. An error is
+// returned if the connection could not be created.
+func CreateVPNSiteConnection(t *testing.T, client *gophercloud.ServiceClient, ikePolicyID string, ipsecPolicyID string, serviceID string, peerEndpointGroupID string, localEndpointGroupID string) (*siteconnections.Connection, error) {
+	t.Helper()
+
+	connectionName := tools.RandomString("ACPTTEST", 16)
+	peerAddress := "172.24.4.233"
```
vooon: Maybe better to allow override that by env var (and default to that ip)?

nicholaskuechler (approved):

integration/networking_test.go
```diff
@@ -1,91 +1,261 @@
...
+func TestNetworkingNetworkMTUCreateUpdateDeleteUpdatesExporterMetrics(t *testing.T) {
+	clients.RequireLong(t)
+
+	networkClient := funcs.NewNetworkClient(t)
+
+	ctx, cancel := funcs.RequestContext(t)
+	defer cancel()
+
+	if _, err := extensions.Get(ctx, networkClient, "net-mtu-writable").Extract(); err != nil {
+		t.Skipf("Neutron net-mtu-writable extension is not available: %v", err)
```
vooon: Similar as for vpnaas - skip only if it's real 404, but not auth or some other weird error.

integration/networking_test.go
```diff
@@ -1,91 +1,261 @@
...
+	scrapeMetrics(t, "after MTU network create").requireLabelValue(t, "openstack_neutron_network", labels{"id": createdNetwork.ID}, "mtu", "1440")
+
+	updateOpts := mtu.UpdateOptsExt{UpdateOptsBuilder: networks.UpdateOpts{}, MTU: 1350}
+	var updatedNetwork funcs.NetworkWithMTU
+	if err := networks.Update(ctx, networkClient, createdNetwork.ID, updateOpts).ExtractInto(&updatedNetwork); err != nil {
+		t.Fatalf("Failed to update Neutron network MTU: %v", err)
+	}
+	scrapeMetrics(t, "after MTU network update").requireLabelValue(t, "openstack_neutron_network", labels{"id": createdNetwork.ID}, "mtu", "1350")
+
+	if err := networks.Delete(ctx, networkClient, createdNetwork.ID).ExtractErr(); err != nil {
```
vooon: Note that with a single RequestContext() all this calls should complete within 60s.
Probably it's ok, just may be a result of a bit flaky CI...

vooon (approved):
Maybe it worth to also poll for some time to obtain expected metric value (not just expected name)?
But overall, i think it's LGTM.

PR #536, opened 2026-03-04
feat: add support for additional labels on metrics

rajsameer:
Add generic --extra-labels flag for customising metric labels

  ## Summary

  Introduces a new --extra-labels CLI flag that allows operators to attach additional labels to any metric across all OpenStack service exporters, without requiring code changes. This
  replaces the narrower, Ironic-specific --ironic.additional-labels approach with a generic solution.

  ## Motivation

  Users often need to enrich metrics with environment-specific context (e.g. rack ID, datacenter, environment name) that is available in OpenStack API responses but not exposed as
  Prometheus labels by default. Previously, adding such labels required service-specific code changes per exporter.

  ## Changes

  New --extra-labels flag (main.go, utils/kingping_extra_labels_parser.go)

  - Accepts one or more --extra-labels=<service>.<metric>:<labels> arguments (cumulative flag)
  - Supports two label types per invocation:
    - Dynamic labels — resolved from API response fields via dot-path (e.g. conductor, extra.rack_id → label name extra_rack_id)
    - Static/const labels — fixed key=value pairs appended to every metric sample (e.g. env=production)
  - Full validation: Prometheus label name constraints enforced, duplicates rejected
  - Example: --extra-labels=ironic.node:conductor,extra.rack_id,datacenter=dc1

  ## Generic reflect-based field resolver (exporters/fields_cahce.go)

  - resolveField(obj, "extra.rack_id") — resolves dot-paths against any struct using a JSON-tag-keyed index; supports nested map[string]interface{} access
  - getFieldIndex — lazily built, thread-safe cache (sync.RWMutex with double-checked locking) keyed by reflect.Type, shared globally across all exporters
  - computeMetricLabels / computeConstantLabels — helpers used by all exporter constructors to append extra variable/const labels to metric descriptors

  All service exporters updated

  Extra label support added to all 15 exporters: ironic, cinder, nova, neutron, glance, keystone, loadbalancer, heat, placement, manila, designate, trove, gnocchi, containerinfra,
  objectstore

  ## Testing

  - utils/kingpin_parser_test.go — 10 new tests covering parsing, dot-path conversion, static labels, cumulative behaviour, error cases, and Extract
  - exporters/fields_cache_test.go — 21 new tests covering resolveField, computeMetricLabels, computeConstantLabels, and resolveExtraLabelValues
  - exporters/ironic_test.go — End-to-end integration test: creates an Ironic exporter with ironic.node:conductor,env=test and validates that all 5 fixture nodes carry both the dynamic
[...]

PR #570, opened 2026-06-15
Refactor metric filtering into DAG-based exporter collection

### LINKED ISSUE #578: Allow selected Ironic driver_info and instance_info fields to be exported as metrics
## Usecase
In production Ironic environments, some fields in driver_info or instance_info are operationally useful to monitor.

One example is temporary deploy ramdisk/kernel overrides. Operators may set driver_info[deploy_ramdisk] or driver_info[deploy_kernel] during break-glass debugging or development testing. These overrides are not intended to remain on production nodes indefinitely, but they can be forgotten.

If these selected fields were available in Prometheus, operators could alert on unexpected or stale values.

## Idea

It would be useful for openstack-exporter to support exporting a small, operator-configured allowlist of Ironic node driver_info and instance_info keys.

For example, an operator may want to expose selected fields such as:

```bash
driver_info[deploy_ramdisk]
driver_info[deploy_kernel]
instance_info[image_source]
```

It may also be worth considering an optional regex allow/deny filter for exported values, so operators can prevent sensitive values from being exposed even when a key is selected.

vooon:
## Summary

Refactor exporter collection from independent per-metric callbacks into a dependency DAG of data sources and metric emitters.

This fixes disabled-metric crashes and makes metric filtering affect the whole collection path: disabled metrics no longer leave emitters or API fetches running with nil descriptors.

## What Changed

- Added typed descriptor declarations and centralized metric policy resolution.
- Replaced callback maps with per-exporter source/emitter DAGs.
- Migrated all exporters to the DAG collection model.
- Prune emitters and their unused transitive sources when metrics are disabled.
- Run independent sources and ready emitters concurrently.
- Added bounded parallel detail fetching for Cinder, Nova, Neutron, Placement, and Designate.
- Added/configured:
  - `--dns-concurrent-count`
  - `--api-detail-concurrent-count`
  - `--placement-concurrent-count`
- `--collect-metric-time` now exposes source-level timing:
  `openstack_exporter_source_fetch_duration_seconds{service,source}`.
- Standalone and probe registries reuse exporters rather than rebuilding clients, descriptors, and microversion setup on every scrape.
- Moved exporter internals, DAG model, and contributor guidance into `docs/exporter.md`.
- Added Vault password lookup documentation.

## Metric Policy And Pruning

Metric enablement is resolved once before descriptor creation and DAG pruning:

1. `--disable-metric` disables a metric absolutely.
2. `--enable-metric` overrides slow/deprecated filtering.
3. `--disable-slow-metrics` disables metrics tagged as slow.
4. `--disable-deprecated-metrics` disables deprecated metrics.
5. Remaining metrics are enabled.

Both descriptor registration and `PruneSchedule` consume this same result.

This specifically fixes the review finding where slow/deprecated descriptors were skipped but their expensive sources still ran. For example, disabling slow metrics now removes Cinder/Nova project limit fetches together with their project dependency, as well as Neutron quota, Glance detail, and Nova usage sources.

Emitter metric names are also validated during construction, preventing a typo from silently keeping an emitter and its sources alive.

## Ironic Node Labels

Closes #578.

`--ironic.node-extra-labels` maps selected top-level keys from Ironic node dictionaries onto `openstack_ironic_node` labels:

```text
driver_info
instance_info
extra
properties
driver_internal_info
```

Use a driver-specific BMC address mapping:
```
# Redfish
--ironic.node-extra-labels=driver_info.deploy_kernel,driver_info.deploy_ramdisk,bmc_address=driver_info.redfish_address

# IPMI
--ironic.node-extra-labels=driver_info.deploy_kernel,driver_info.deploy_ramdisk,bmc_address=driver_info.ipmi_address
```

*bmc_address must appear only once per mapping; choose the key matching the driver in use.*

The default remains equivalent to the old output: `driver_info.deploy_kernel,driver_info.deploy_ramdisk`
Pass an empty value to remove those labels.

Credential-like mapping keys are warned about and their values are redacted.
This covers keys ending in password, passwd, secret, token, or private_key, plus configdrive and image_url.

Ironic node collection now uses deterministic marker pagination with `limit=1000`, `sort_key=id`, and `sort_dir=asc`. This **fixes incomplete collection** on deployments with more than 1,000 nodes.


niedbalski (changes requested):
Not to discard this other than to get context, but I am opposed to adding super extra complexity for not a extremely good reason. What is the rationale for this change?

vooon:
@niedbalski it's actually quite simple, `.Fn` approach only good working for simple metrics, without cross-dependencies, but there already plenty of places where you either fetch the same list twice, or just place everything into a single `Fn`. Then you have `--disable-metric`, which may drop some of the metrics from `exporter.Metrics`, and those monolithic `Fn`s became a SIGSEGV source.
To fix that you have to go trough whole codebase and add a nil check. Not something i'd like to support. Instead a simple dependency tree manager just makes that simpler, and probably may allow having more parallelism in fetch later.

Dropping `exporter.Metrics` mapping in favor of struct also makes the code simpler, faster and compile-time verifiable (two pointer refs always faster than map lookup, and the search key always known).

vooon:
@niedbalski to elaborate further, what would happen if i provide `--disable-metric placement-resource_provider_allocations` (1), and what `--disable-metric placement-resource_total` (2).
What if i add `--disable-metric placement-resource_usage` (3)?

1. Nothing, because i have added explicit test for that particular metric: https://github.com/openstack-exporter/openstack-exporter/blob/518cc56334e5bd1b73f0f24d0a9a83cc69bd58e2/exporters/placement.go#L80
2. Would SEGFAULT at https://github.com/openstack-exporter/openstack-exporter/blob/518cc56334e5bd1b73f0f24d0a9a83cc69bd58e2/exporters/placement.go#L114
3. Also SEGFAULT, but more important that before that it'll call `GetUsages()`, while i completely do not need it.

So a quite simple graph scheduler centralise all that conditions, so now it less possible to make a mistake.

vooon:
Interesting speedup on our Ironic installation like a 10x:

Screenshot 2026-06-24 at 09-02-00 Explore - Core Openstack - Grafana
Screenshot 2026-06-24 at 09-08-37 Explore - Core Openstack - Grafana

vooon:
@Sharpz7 well, that's yet another huge PR... It's now on testing on all our clouds (and so far, looks good)...
Please try copilot review.

Sharpz7:
Hey @vooon I think I'll try reviewing this one with my AI tools locally. Will report back anything useful. Thank you for this!

exporters/exporter.go
```diff
@@ -19,217 +23,776 @@ import (
...
-	if exporter.CollectTime {
-		ch <- prometheus.MustNewConstMetric(exporter.Metrics["openstack_metric_collect_seconds"].Metric, prometheus.GaugeValue, time.Since(now).Seconds(), metricName)
+// PruneSchedule computes a schedule containing only nodes needed given base's
+// enabled/disabled metric configuration. An emitter is dropped when all of its
+// declared metrics are disabled; a source is dropped when no live emitter
+// (transitively) depends on it. Call once at construction time.
+func (g *Graph[E, S]) PruneSchedule(base *BaseOpenStackExporter) (Schedule, error) {
+	liveEmit := make([]bool, len(g.Emitters))
+	for i, em := range g.Emitters {
+		for _, m := range em.Metrics {
```
Sharpz7, line 381: I think this does not fully prune the DAG for `--disable-slow-metrics` or `--disable-deprecated-metrics`. `RegisterAndFillDescs` skips those descriptors, but `PruneSchedule` only asks `IsMetricEnabled`, and that currently only checks `DisabledMetrics`. That means expensive slow/deprecated sources can still run, just with nil descriptors/no emitted samples. This undercuts the pruning/performance intent, especially for project/detail sources.

main.go
```diff
@@ -258,8 +280,14 @@ func startHTTPServer(ctx context.Context, services []string, toolkitFlags *web.F
...
 	} else {
-		logger.Info("openstack exporter started in legacy mode")
-		http.HandleFunc(*metrics, metricHandler(services, logger))
+		logger.Info("openstack exporter started in standalone mode")
+		h, err := buildStandaloneHandler(services, opts, logger)
+		if err != nil {
+			logger.Error("Failed to build exporters", "error", err)
+			cancel(err)
+			return
+		}
```
Sharpz7, line 295: This registry/exporter reuse looks like the most likely explanation for a large reported baremetal speedup. The Ironic exporter still performs a single `nodes.ListDetail(...).AllPages()` source, so the Ironic API work itself is not substantially parallelized here. What changes is that repeated scrapes no longer rebuild registries/exporters/clients/descriptors/microversion setup on every request. If the benchmark included that setup path, baremetal could look dramatically faster even though the `/nodes/detail` call is essentially unchanged.

docs/metrics.md
```diff
@@ -0,0 +1,1475 @@
+# Metrics
```
Sharpz7: Can we split the generated docs work into a separate PR? This file alone adds about 1.5k lines, and the generator adds another 600+ lines, which makes the behavioral/performance refactor much harder to review. A docs-only follow-up would keep this PR focused on the exporter/DAG changes.

Sharpz7 (commented):
Leaving a few review notes from a performance/size pass, especially around the reported baremetal speedup and PR scope.

Assistance from: Codex GPT-5.5

main.go
vooon: The speedup not tied directly to ironic api, more like to general parallelisation and faster placement collection.

docs/metrics.md
vooon: Ok, can do.

exporters/exporter.go
vooon: Will check.

Sharpz7:
Hey @vooon. Sorry, I am not sure I have a better way to reach you than here.

We have an openstack-exporter channel in https://docs.openstack.org/contributors/es_MX/common/irc.html that we are trying to re-vitalize. It would be great to have you there?

Also, I am using the [needs-more-reviews](https://github.com/openstack-exporter/openstack-exporter/pulls?q=is%3Aopen+is%3Apr+label%3Aneeds-more-reviews) label for PR's that I have authored, and need someone elses approval for me to feel comfortable merging. If you are able to help out with that, it would be great. I think PR's that have the small label make sense to do first?

vooon:
Oh, IRC, haven't used it for ages... :)
My bandwidth varies, don't hesitate to ping me.

docs/metrics.md
vooon: Moved away.

exporters/exporter.go
vooon: Refactored.

vooon:
@Sharpz7 i've done huge rebase, some refactoring, done #578, and also found and fixed the bug with >1000 ironic nodes fetching.

Sharpz7:
I think this PR needs to be split before we can review it properly. My suggestion would be five PRs: the core DAG/metric-filtering refactor; exporter and registry reuse; concurrent API collection; configurable Ironic labels; and the Ironic pagination fix. I do not think it needs to be split any further than that.

I would particularly like to understand the reported 10× performance improvement before merging the concurrency PR. At the moment, it is attributed to general parallelization and faster Placement collection, but registry reuse may also explain some of it. Splitting these changes should make this relatively easy to test. The Placement results should also be compared directly with #546, including its generation-based cache.

The Ironic labels overlap with #536, so we should decide whether this is intended to be a narrower but better solution, or whether the generic approach in #536 is preferable. I also wonder whether fixed security redactions are enough. Sensitive fields may have deployment-specific names, so allowing custom redaction rules or regexes may be more useful.

Sharpz7 (changes requested):

vooon:
@Sharpz7 ok, i will do new PRs, one at a time. First one  - to introduce DAG engine + port the plugins (i do not want to make every commit buildable - so not to keep old engine and lots of messing around).

PR #573, opened 2026-06-20
fix(nova): handle nil CPUInfo in ListHypervisors to prevent running_vms collection panic

### LINKED ISSUE #571: running_vms metric missing: unhandled nil CPUInfo panic on SHUTOFF instances (Nova 2026.1)
**Reported By**
**Investigated by: Sailappay Mohammed Zuhaib**
**Reviewed and Authorized by: Bhavana S**
**Organisation: Centre for Development of Advanced Computing (C-DAC)**

**Environment**

| Field | Value |
|---|---|
| Exporter image | `quay.io/openstack.kolla/prometheus-openstack-exporter:2026.1-ubuntu-noble` |
| Exporter build | `go=go1.19.13, platform=linux/amd64` |
| OpenStack release | 2026.1 (Flamingo) |
| Nova version | `33.0.2.dev9` |
| Deployment | Kolla-Ansible 2026.1, all-in-one |
| OS | Ubuntu Noble 24.04, kernel `6.17.0-1017-aws` |
| Exporter config | `OS_COMPUTE_API_VERSION=latest` |

**Bug Description**

The `running_vms` metric collector fails completely and permanently
when any Nova instance is in **SHUTOFF** (powered-off) state.

The exporter iterates all Nova instances and calls
`GET /v2.1/servers/{uuid}/diagnostics` for each one.
When a SHUTOFF instance is encountered, Nova correctly returns
**HTTP 409 Conflict** — this is expected, documented Nova behavior.

The exporter **does not handle this 409 response**. It panics on
a nil type assertion for the `CPUInfo` field:

error: CPUInfo has unexpected type: <nil>

This crashes the **entire** `running_vms` collection loop — not just
the single SHUTOFF instance. The result is that
`openstack_nova_running_vms` is **completely absent** from the
exporter output as long as any VM is SHUTOFF.

The error fires on **every single 60-second scrape cycle**
with no recovery.

**Steps to Reproduce**
Step 1
**1.** Have at least one Nova instance in SHUTOFF state:
```bash
openstack server stop <any-server>

**2.** Wait 60 seconds, then check exporter logs:
docker logs --since=2m prometheus_openstack_exporter 2>&1 | grep CPUInfo

**3.** Confirm metric is missing from exporter output:
curl -s http://EXPORTER_HOST:9198/metrics | grep 'running_vms'
# returns empty — metric completely absent

**4.** Confirm Nova returns 409 for SHUTOFF instance:
curl -s -H "X-Auth-Token: TOKEN" \
  http://NOVA_HOST:8774/v2.1/servers/SHUTOFF_VM_UUID/diagnostics

**Actual Result**

Step 2 — Exporter log (fires every 60 seconds, indefinitely):
ts=2026-06-17T22:16:56.009Z caller=exporter.go:100
level=info msg="Collecting metrics for exporter"
exporter=openstack_nova metrics=running_vms

ts=2026-06-17T22:16:56.009Z caller=exporter.go:126
level=error err="Failed to collect metric for exporter"
exporter=nova
error="failed to collect metric: running_vms,
error: CPUInfo has unexpected type: <nil>"

Step 3 — Exporter /metrics output:
# running_vms metric is completely absent
# openstack_nova_total_vms 5   ← this still works (different code path)
Step 4 — Nova API response for SHUTOFF instance:

{
  "conflictingRequest": {
    "code": 409,
    "message": "Cannot 'get_diagnostics' instance
    41411f01-f907-4326-8186-eff82510af66
    while it is in power state shutdown"
  }
}

**Expected Result**
The running_vms collector should:

Check the HTTP response code before parsing the response body
On receiving HTTP 409 — skip that instance gracefully
Continue collecting metrics for all other instances
Export openstack_nova_running_vms correctly for active VMs

**Root Cause**
Nova API correctly returns 409 for SHUTOFF instances. This is not a Nova bug.

The bug is in the exporter's Go code in the running_vms collector:

Instance loop hits a SHUTOFF VM
Calls /diagnostics → receives 409
CPUInfo field is nil in the parsed response
Type assertion on nil → panic / collection failure
Entire loop aborts — not just this one instance

A related nil fix was made in commit 7533071 for nova_server_status. The same pattern needs to be applied to the running_vms collector.

**Impact**
openstack_nova_running_vms permanently missing from Prometheus
Affects any deployment with at least one stopped VM
Core operational metric silently gone — no alert fires
openstack_nova_total_vms shows 5 but running_vms is absent — discrepancy is hard to detect without knowing about this bug

**Workaround**
Pin Nova microversion in clouds.yaml to prevent the modern diagnostics API path:

clouds:
  default:
    compute_api_version: "2.35"

Reproduced on: Kolla-Ansible 2026.1, Nova 33.0.2, June 2026 Also reported on Launchpad: link

### PR DESCRIPTION
**Pull request by smzuhaib06**
## Summary
Fixes a nil panic in `ListHypervisors` that causes `openstack_nova_running_vms`
to be completely absent from Prometheus when any hypervisor returns `nil` for `CPUInfo`.

## Problem
When Nova returns `nil` for `CPUInfo` (occurs with certain hypervisor states
or API microversions), the existing `reflect.ValueOf(hypervisor.CPUInfo).IsZero()`
check does not safely guard against nil pointer access on the `Topology` struct.

This aborts the entire collection loop with:
error: CPUInfo has unexpected type: <nil>

The error fires on **every 60-second scrape cycle** with no recovery,
permanently removing `openstack_nova_running_vms` from Prometheus output.

## Root Cause
`hypervisor.CPUInfo` is a pointer/interface type.
`reflect.ValueOf(nil).IsZero()` does not return `true` for nil interfaces —
it can still panic when accessing `.Topology` fields on the nil value.

## Fix
Extract `CPUInfo` to a local variable and check:
1. Not nil
2. Not IsNil() (for interface types)
3. Topology fields are non-zero before multiplication

Fall back to `hypervisor.VCPUs` when `CPUInfo` is unavailable.

## Testing
Reproduced and verified on:
- Kolla-Ansible 2026.1, all-in-one
- Nova 33.0.2.dev9
- Exporter image: `quay.io/openstack.kolla/prometheus-openstack-exporter:2026.1-ubuntu-noble`

Before fix:
level=error error="CPUInfo has unexpected type: <nil>" openstack_nova_running_vms = absent

After fix (with workaround applied):
openstack_nova_running_vms = present and correct

## Related
- Closes #571 
- Similar nil fix was done in commit 7533071 for `nova_server_status`
- Also reported on Launchpad: https://bugs.launchpad.net/kolla-ansible/+bug/2157070

### CODE REVIEW

**Comment by smzuhaib06** [comment]
### Summary of Changes
- Extracted `CPUInfo` to a local variable and implemented a direct, type-safe boundary check on inner topology integers (`Cores`, `Threads`, `Sockets`). 
- Removes the flaky `reflect` checks that caused fatal `<nil>` unmarshaling panics on down/SHUTOFF hypervisors.
- Added a regression test case in `nova_test.go` covering an empty `CPUInfo{}` payload to ensure safe fallback to default `VCPUs`.
- Verified all local unit tests pass cleanly.

Closes #571. Ready for review!

**Comment by phsm** [comment]
I concur, I also see this bug in my Openstack installation

File: exporters/nova.go [hunk 1]
```diff
@@ -224,8 +223,13 @@ func ListHypervisors(ctx context.Context, exporter *BaseOpenStackExporter, ch ch
 			prometheus.GaugeValue, float64(hypervisor.CurrentWorkload), hypervisor.HypervisorHostname, availabilityZone, aggregates)
 
 		var vcpus int
-		if !reflect.ValueOf(hypervisor.CPUInfo).IsZero() {
-			vcpus = max(hypervisor.CPUInfo.Topology.Cells, 1) * hypervisor.CPUInfo.Topology.Sockets * hypervisor.CPUInfo.Topology.Cores * hypervisor.CPUInfo.Topology.Threads
+		cpuInfo := hypervisor.CPUInfo
+		// Check the values of the Topology fields instead of checking the struct for nil
```
Sharpz7, line 225: `hypervisor.CPUInfo` is a value `hypervisors.CPUInfo` in the pinned Gophercloud v2.13.0, not a pointer or interface. Its unmarshaller leaves the zero value when `cpu_info` is missing or `null`, and the existing `reflect.ValueOf(...).IsZero()` check already detects that and falls back to `hypervisor.VCPUs` without a panic. Consequently, this test also passes against the pre-change code, so it does not demonstrate the reported regression. Could you provide a failing response/stack trace for the version in `go.mod`, or narrow this PR to the distinct change in behavior for a present but incomplete topology?

Sharpz7 (commented):
Found one blocking correctness issue: the proposed nil-panic diagnosis is not supported by the pinned Gophercloud CPUInfo type, and the new test does not fail before this change.

Assisted-By: Codex

Sharpz7:
If the above is more of a nit than a proper problem, let me know and I can just approve and merge

Issue #491, opened 2025-09-30
openstack_nova_running_vms metrics erroring out


vipin5050:
I am using openstack exporter for prometheus https://github.com/openstack-exporter/openstack-exporter but I am not getting openstack_nova_running_vms metrics and I see an error message in logs

`time="2025-09-25T21:36:07Z" level=error msg="Failed to collect metric for exporter: nova, error: failed to collect metric: running_vms, error: CPUInfo has unexpected type: <nil>" source="exporter.go:123"`

I checked 
#openstack hypervisor show <hostname> and indeed cpu_info is None 

`| cpu_info | None `

my nove.conf file and I see cpu_mode is defiened as passthrough and we can not change it as we need that. 
looks like code is coming from https://github.com/openstack-exporter/openstack-exporter/blob/main/exporters/nova.go around line 224, where it bails our as in my case CPU_Info is None. In my view code should gracefully handle if cpu_info comes as nil or None instead of erroring and not provide any metrics for running vms.

Sharpz7:
We are aware of this issue, and the potential for it to happen in the future. Tracked in https://github.com/openstack-exporter/openstack-exporter/issues/592

Issue #470, opened 2025-07-31
OpenStack Exporter on RHEL 8.10 with OpenStack 17: "Resource not found" errors for security groups metrics


noor47-pixel:
**Description:**
I am running the OpenStack exporter in a container using Podman on RHEL 8.10. The exporter is configured with a `clouds.yaml` file to connect to a Red Hat OpenStack Platform 17 deployment.

Since upgrading to OpenStack 17, the exporter struggles to retrieve some information related to security groups. I repeatedly see the following error in the logs:

```
ts=2025-05-07T15:51:20.362Z caller=main.go:169 level=info msg="Starting openstack exporter version for cloud" version="(version=, branch=, revision=1a2e1138ca951da736a974cbeba91fda4501afb0-modified)" cloud=openstack
ts=2025-05-07T15:51:20.363Z caller=exporter.go:205 level=info msg="SSL verification disabled on transport"
…
ts=2025-05-07T15:51:34.639Z caller=exporter.go:100 level=info msg="Collecting metrics for exporter" exporter=openstack_nova metrics=security_groups
ts=2025-05-07T15:51:34.643Z caller=exporter.go:126 level=error err="Failed to collect metric for exporter" exporter=nova error="failed to collect metric: security_groups, error: Resource not found: [GET https://url_of_openstackv17:13774/v2.1/os-security-groups], error message: {\"itemNotFound\": {\"code\": 404, \"message\": \"The resource could not be found.\"}}"
```

This leads to latency issues and repeated errors in the logs.

I suspect that since I upgraded OpenStack to version 17, some API endpoints (like `/v2.1/os-security-groups`) might no longer be available or have changed.

* I run the exporter as a container with Podman and provide `clouds.yaml` for authentication.
* SSL verification is disabled as noted in the logs.
* Could the API endpoint have changed in OpenStack 17?
* Is there a recommended approach to update the exporter or configuration for this OpenStack version?

Any guidance or similar experiences would be appreciated.

Issue #221, opened 2022-02-08
openstack_nova_security_groups metric unavailable 


ici0:
exporter errored with
`ERRO[0036] Failed to collect metric for exporter: nova, error: failed to collect metric: security_groups, error: Resource not found: [GET https://my.domain.cloud:8774/v2.1/os-security-groups], error message: {"itemNotFound": {"code": 404, "message": "The resource could not be found."}}  source="exporter.go:123"
`
nova droped '/os-security-group*' since 2.36 api microversion
https://docs.openstack.org/nova/wallaby/reference/api-microversion-history.html#id33 

maybe openstack exporter should drop the metric too

nova: wallaby
openstack-exporter: 1.4.0, master

Hybrid512:
Not pretty sure that most of the Openstack clusters installed out there are all >=wallaby.
Would be better to return some null value (0, -1, ... anything else) in case of error from gophercloud instead of throwing an error of that kind.
Or maybe support Openstack versioning and deprecations in the API but it would mean a lot of work and this is useless since the exporter rely on gophercloud which should do it already.

priteau:
Please note that you can set `OS_COMPUTE_API_VERSION` as an environment variable to get this metric to work again.

mstinsky:
> Please note that you can set `OS_COMPUTE_API_VERSION` as an environment variable to get this metric to work again.

Another side node for this. Pinning the microversion is not possible in this case as the api was completely removed with openstack ussuri.

gtherond:
I actually use ZED and put the microversion using the environment variable, but as openstack API gives a way to discover available and supported version, we shouldn't have to do that when looking to target the currently latest available version on the targeted cloud.

We should use that feature only when targeting either OLD/DEPRECATED or EDGE version that aren't yet considered as prod version.

Sharpz7:
https://github.com/openstack-exporter/openstack-exporter/issues/592

We will try and address issues like this long term going forward

PR #581, opened 2026-07-02
Microversion overhaul

Sharpz7:
## Summary

- overhaul microversion setup so services that advertise microversion ranges are configured centrally
- remove the hardcoded Nova `2.87` cap and use Nova discovery by default when `OS_COMPUTE_API_VERSION` is not set
- keep explicit env overrides strict: unsupported requested microversions fail instead of silently collecting misleading metrics
- document supported `OS_*_API_VERSION` variables in `docs/microversions.md`
- add a README documentation entry point for the new docs folder

## Issues this may solve

- #491: uses the newest Nova compute API microversion advertised by the cloud unless pinned by `OS_COMPUTE_API_VERSION`
- #249: supports Wallaby-era Nova/Placement behavior by removing the Nova microversion cap and documenting compute microversion handling
- #221: may restore Nova security group metrics when the failure is caused by using an older compute API microversion
- #220: may restore flavor ID collection for `nova_server_status` when the field depends on a newer compute API microversion
- #470: may help if the RHEL/OpenStack 17 security group errors are caused by Nova API version availability, but this still needs deployment validation

## Validation

- `git diff --check`
- `GOCACHE=/tmp/openstack-exporter-go-cache go test ./utils ./exporters`
- `GOCACHE=/tmp/openstack-exporter-go-cache go test ./...` (fails only on integration tests requiring OpenStack env/clouds.yaml: missing `OS_AUTH_URL` and clouds config)


Sharpz7:
Closing this. We can redo it based on where https://github.com/gophercloud/gophercloud/issues/3882 goes

Issue #587, opened 2026-07-21
Critical/High CVEs in released docker images


dmbuck32:
## Overview
We've added trivy scanning to our internal CI pipelines and noticed that the docker image from the latest release (`v1.8.0-alpha`) contains a couple of critical CVEs. Additionally, the `latest` image contains a couple of high CVEs that should be fixed by bumping the pinned go version.

### Scan Results

<details>
<summary>v1.8.0-alpha</summary>

```
trivy image --severity CRITICAL,HIGH ghcr.io/openstack-exporter/openstack-exporter:1.8.0-alpha
2026-07-22T10:58:29+12:00	INFO	Vulnerability scanning is enabled
2026-07-22T10:58:29+12:00	INFO	Secret scanning is enabled
2026-07-22T10:58:29+12:00	INFO	If your scanning is slow, please try '--scanners vuln' to disable secret scanning
2026-07-22T10:58:29+12:00	INFO	Please see also https://aquasecurity.github.io/trivy/v0.52/docs/scanner/secret/#recommendation for faster secret detection
2026-07-22T10:58:30+12:00	INFO	Detected OS	family="debian" version="12.12"
2026-07-22T10:58:30+12:00	INFO	[debian] Detecting vulnerabilities...	os_version="12" pkg_num=6
2026-07-22T10:58:30+12:00	INFO	Number of language-specific files	num=1
2026-07-22T10:58:30+12:00	INFO	[gobinary] Detecting vulnerabilities...

ghcr.io/openstack-exporter/openstack-exporter:1.8.0-alpha (debian 12.12)

Total: 8 (HIGH: 7, CRITICAL: 1)

┌─────────┬────────────────┬──────────┬────────┬───────────────────┬──────────────────┬──────────────────────────────────────────────────────────────┐
│ Library │ Vulnerability  │ Severity │ Status │ Installed Version │  Fixed Version   │                            Title                             │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼──────────────────┼──────────────────────────────────────────────────────────────┤
│ libssl3 │ CVE-2026-31789 │ CRITICAL │ fixed  │ 3.0.17-1~deb12u3  │ 3.0.19-1~deb12u2 │ openssl: OpenSSL: Heap buffer overflow on 32-bit systems     │
│         │                │          │        │                   │                  │ from large X.509 certificate...                              │
[...]

PR #589, opened 2026-07-21
Add Trivy Vulnerability Reporting to CI pipeline.

dmbuck32:
Add Trivy Vulnerability Reporting the CI pipeline.

Should help identify issues like #587.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -125,6 +125,35 @@ jobs:
           build-args: |
             GO_VERSION=${{ steps.go-version.outputs.version }}
 
+      # The first call to the action will invoke setup-trivy and install trivy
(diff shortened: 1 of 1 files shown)


Sharpz7:
@dmbuck32 is there a reason to pick this over Github's default vulnerability scanner? Its currently disabled for this repo but I can re-enable it?

dmbuck32:
> @dmbuck32 is there a reason to pick this over Github's default vulnerability scanner? Its currently disabled for this repo but I can re-enable it?

I've not used Github's default vulnerability scanner previously. I don't have an opinion on whether to use trivy or the default scanner.

Sharpz7:
@dmbuck32 in that case, I think I will enable the default. Thank you for bringing this to my attention.

PR #593, opened 2026-07-28
Extend glance metrics to have image type property

iaas-aannigeri:
## What

Adds an `image_type` label to `openstack_glance_image_bytes` and `openstack_glance_image_created_at`, sourced from the `image_type` image property that nova sets on server snapshots (`snapshot`) and server backups (`backup`). Images without the property (regular uploads) are labeled `image`.

## Why

Today the glance metrics make it impossible to tell a plain image from a VM snapshot or VM backup, so operators cannot break down glance storage usage by kind, e.g.:

```
sum by (image_type, tenant_id) (openstack_glance_image_bytes)
```

Example of a nova server snapshot as reported by the API (trimmed):
```json
{
  "name": "test-vm-snapshot",
  "properties": { "image_type": "snapshot", "instance_uuid": "2d1029d0-..." }
}
```

## Notes

- `image_type` values are passed through as-is, so nova's `backup` (from `nova backup") and any other values are preserved; only the absent case is normalized to `image`.
- Fixture, expected test output and README updated.
- Verified against a live cloud (2 server snapshots + 28 plain images reported correctly).


Sharpz7:
It would be great if you could add integration tests for these PR's.

The integration tests live in [the `integration/` directory](https://github.com/openstack-exporter/openstack-exporter/tree/main/integration).

iaas-aannigeri:
> It would be great if you could add integration tests for these PR's.
> 
> The integration tests live in [the `integration/` directory](https://github.com/openstack-exporter/openstack-exporter/tree/main/integration).

Tests are in place :)

Sharpz7 (approved):

PR #563, opened 2026-06-02
feat: add ironic node labels

Sharpz7:

PR #596, opened 2026-08-02
Add optional Ironic driver_info labels

### LINKED ISSUE #578: Allow selected Ironic driver_info and instance_info fields to be exported as metrics

Sharpz7:
Adds an opt-in flag to expose Ironic deploy_kernel and deploy_ramdisk fields as node metric labels.

Closes #578


nicholaskuechler (approved):

Sharpz7:
@vooon I think one of your upcoming changes contains something that superceeds this?

vooon:
@Sharpz7 yes, i have a patch that allows to expose any of properties as a labels. And setting that mapping to `""` will hide this two driver info.

Sharpz7:
Current upstream main already exports `deploy_kernel` and `deploy_ramdisk` on `openstack_ironic_node` via #563, so this PR is redundant for those fields. The broader #578 allowlist remains separate because `driver_info` and `instance_info` may expose secrets; Voon is working on that direction.

Sharpz7:
I still think we should keep this PR's direction of these things being opt-in. We can discuss more soon

PR #610, opened 2026-08-06
fix: remove internal Ironic self-healing label

Sharpz7:
## What changed
Removed `ironic_self_healing_state` from the `openstack_ironic_node` metric labels.

## Why
The value is an internal Ironic `extra` field and should not be exported as a Prometheus label.

## Validation
- `go test ./exporters`

The Ironic fixture retains the internal field, while the expected metric output omits it, preventing regression.


Sharpz7:
@vooon - this is an internal metric that needs removed. A quick approval would be great

vooon (approved):
You could have kept it in fixture, but okay.
With mapping like i have done user can return it, if he really needs it...

Sharpz7:
@vooon eh, it was discussed in IRC to nuke it. I will review your change when I can

PR #616, opened 2026-08-15
fix: collect all Ironic nodes on large deployments

vooon:
Split out of #570, as requested, so it can be reviewed and merged on its own. This is the Ironic pagination fix; the other four parts (core DAG/metric-filtering refactor, exporter and registry reuse, concurrent API collection, configurable Ironic labels) will follow as separate PRs.

## Problem

`ListNodes` calls `nodes.ListDetail(...).AllPages()` with no options. The detailed node endpoint applies a server-side default limit, and it does not always return `nodes_links`, so `AllPages` has no next link to follow and returns only the first slice of the inventory. Deployments with more nodes than that default have been silently reporting a truncated node list — no error, no warning, just missing `openstack_ironic_node` series.

## Fix

Page explicitly: request `limit=1000` under a stable sort (`sort_key=id`, `sort_dir=asc`) and continue from the UUID of the last node whenever a page comes back full. A short page ends collection.

The marker is exclusive, so no node is emitted twice. Where the API *does* return `nodes_links`, `AllPages` still follows them and the extra marker request simply comes back empty.

## Testing

`TestListAllNodesFollowsMarker` serves a full 1000-node first page without `nodes_links` — exactly the shape that truncates today — followed by a 3-node page, and asserts all 1003 nodes are collected in order. Without the marker loop the first page is all you get, so the test fails against the previous behaviour.

The mocked Ironic URL in the exporter test fixtures gains the new query string; the remaining churn in that file is `gofmt` realigning the map.

Made with [Cursor](https://cursor.com)


vooon:
@Sharpz7 sorry to be pushy, but that one blocks me from next #570 split, can it be merged?

exporters/ironic.go
```diff
@@ -58,14 +67,44 @@ func NewIronicExporter(config *ExporterConfig, logger *slog.Logger) (*IronicExpo
...
+		if err != nil {
+			return nil, err
+		}
+
+		page, err := nodes.ExtractNodes(allPagesNodes)
+		if err != nil {
+			return nil, err
+		}
+
+		allNodes = append(allNodes, page...)
```
Sharpz7, line 93: Could we keep paging until an empty page? `max_limit` can be configured below 1000, so a short page may still have more nodes.

Sharpz7 (commented):
I found one pagination edge case to address.

Assisted-By: Codex

Sharpz7:
@vooon also these PR's not being merged probably doesn't stop you building the other PR's on top of this one, in a relation chain?

It will make the diffs look weird to begin with, but that is okay I think.

You could also try https://github.github.com/gh-stack/?

Sharpz7:
Hey @vooon , any update on this?

vooon:
@Sharpz7 sorry, was busy on other projects. I've added fix for max_limit you mentioned.

vooon:
I cannot make stacked PRs:

```
$ gh stack submit
Checking stack state...
? Multiple remotes found. Which remote should be used? gh-vooon
? Save "gh-vooon" as the default remote for all gh stack operations? No
Pushing to gh-vooon...
PR #616 for rebase-570-ironic-limit-1000-fix is up to date
⚠ failed to update base branch for PR #638: HTTP 422: Validation Failed (https://api.github.com/repos/openstack-exporter/openstack-exporter/pulls/638)
PullRequest.base is invalid
⚠ Stacked PRs are not enabled for this repository
✓ Pushed and synced 2 branches
```

Issue #605, opened 2026-08-02
Automatically select the latest supported microversion


Sharpz7:
Use [GetServiceVersions](https://github.com/gophercloud/gophercloud/blob/main/openstack/utils/discovery.go) to select each endpoint’s highest supported version, preserving explicitly configured environment overrides and clear diagnostic logging.

PR #618, opened 2026-08-20
Add regexp-filtered placement provider traits metric

phsm:
This pull request adds support for exporting Placement resource provider traits. 

Motivation behind of of it:
If an Openstack installation consists of multiple groups of hosts (for example, hosts having different CPU model/architecture), and the resource provider traits are used to steer different flavors to different groups of hosts, then there should be a way to determine which traits each host has. 
Knowing that, it will be possible to monitor the capacity of each such group.

Implementation:
Since each resource provider (a compute node) has around 90 traits, it is suggested to use a Regex to only export selected traits.
By default, this regex is set to "CUSTOM_", only exporting the custom traits provided by the operator.

The trait metrics look like this:
```
openstack_placement_resource_provider_trait{hostname="host01",trait="CUSTOM_MY_TRAIT"} 1
openstack_placement_resource_provider_trait{hostname="host01",trait="CUSTOM_MY_OTHER_TRAIT"} 1
openstack_placement_resource_provider_trait{hostname="host02",trait="CUSTOM_MY_TRAIT"} 1
```

Since retrieving traits for every resource can take a lot of time (1 API call per resource), I've marked this metric as "slow". 

Any feedback is welcome.


Sharpz7:
Hey @phsm ! Thank you for this PR. Assuming its not an insane amount of work, could we try and get some integration tests in for this change?

https://github.com/openstack-exporter/openstack-exporter/tree/main/integration

phsm:
I have updated the PR:
- Added missing regex parameter to some `utils_test.go` and `utils.go`
- Added placement integration test. 

Please review and let me know if it misses something.

exporters/placement.go
```diff
@@ -32,6 +34,9 @@ func NewPlacementExporter(config *ExporterConfig, logger *slog.Logger) (*Placeme
 			logger:         logger,
 		},
 	}
```
Sharpz7, line 39: Could we avoid forcing `1.6` on the shared Placement client? Older deployments may reject that microversion, so the existing provider, inventory, usage, and allocation metrics fail too. A separate client for traits or the version handling from #605 would avoid that.

exporters/placement.go
```diff
@@ -61,6 +66,17 @@ func ListPlacementResourceProviders(ctx context.Context, exporter *BaseOpenStack
 			return err
 		}
 
+		if traitsResult, err := resourceproviders.GetTraits(ctx, exporter.ClientV2, resourceprovider.UUID).Extract(); err == nil {
+			for _, v := range traitsResult.Traits {
+				if !exporter.PlacementProviderTraitRegex.MatchString(v) {
```
Sharpz7: Could we skip this when `resource_provider_trait` is disabled? With `--disable-slow-metrics` or `--disable-metric`, its descriptor is absent, so a matching trait reaches this call and can panic on the nil map entry.

README.md
```diff
@@ -507,6 +505,7 @@ openstack_placement_resource_allocation_ratio| hostname="compute-01",resourcetyp
 openstack_placement_resource_generation| hostname="compute-01",resourcetype="DISK_GB\|PCPU\|VCPU\|..."                                                                                                                                                                                                                                                           |20 (float)| Placement resource provider generation for this resource type
```
Sharpz7: **[P2] Fix the documented metric name.** The implementation registers `openstack_placement_resource_provider_trait` (singular), but this table lists `openstack_placement_resource_provider_traits` (plural), so users copying the documented name cannot find the metric.

Sharpz7 (commented):
Could we keep the two Placement fixes from the inline comments? Also, handling Placement microversions through #605 would avoid hardcoding `1.6` here.

Assisted-By: Codex

Sharpz7 (changes requested):
Could we address the two Placement issues in the inline comments before merging? Handling the Placement microversion through #605 would also avoid hardcoding `1.6` here.

Assisted-By: Codex

exporters/placement.go
phsm: I have followed the path of detecting if the microversion is supported. The microversion for shared Placement client is only set if its supported.

exporters/placement.go
phsm: Looks like when a metric is disabled, it ends up not being added to `exporter.Metrics` map. 
So I check whether the the map contains`resource_provider_trait` key before trying to obtain provider traits.

phsm:
I've forgot to mention here that I've updated the PR:

- Added the supported microversions detection. If the microversion 1.6 is supported, it set, and the traits are scraped.
- Added a guard expression: if `exporter.Metrics` map contains `resource_provider_trait` key, it means the metric is not disabled, and the slow metrics are allowed. 

Later I noticed that if the `resource_provider_trait` is disabled, then the microversion was still polled and set without any actual need for it. So I fixed that behavior too with the second commit.

Issue #603, opened 2026-08-02
Add new metrics in bulk with integration tests


Sharpz7:
As a one-off event, I suggest covering all issues in https://github.com/openstack-exporter/openstack-exporter/issues?q=is%3Aopen%20is%3Aissue%20label%3A%22new-metrics%22 in one go, along with adding integration tests.

This would allow us to take down around 10 issues, as we slowly try and get the number of issues down.

Sharpz7:
@vooon what would your thoughts be here? 

cc @niedbalski

vooon:
@Sharpz7 i have reviewed the list, but for many i have no concrete opinion. For some of them i left a comment.

Sharpz7:
PR #624 now covers these concrete metrics issues:

- #203 snapshots by tenant
- #260 Nova task state
- #314 project enabled gauge
- #347 Keystone user info
- #375 security groups on servers and ports
- #473 Nova server groups
- #531 Cinder volume host label

The PR body contains the corresponding Closes directives. #578 already has separate open coverage in #596; #212, #309, and #479 need client/API support; #245 and #317 need broader design decisions; #83 is a broad umbrella issue.

Sharpz7:
Update: PR #624 now covers these eleven issues:

#203, #212, #245, #260, #309, #314, #347, #375, #473, #479, and #531.

The PR body contains a Closes directive for each. #578 is covered separately by PR #596. The only remaining open-list items are #83 and #317, both broad umbrella requests rather than one concrete exporter flow.

Sharpz7:
Update: draft PR #624 now covers these twelve issues:

#83, #203, #212, #245, #260, #309, #314, #347, #375, #473, #479, and #531.

#578 remains covered separately by PR #596. #317 is the only remaining item from this list; it is a broad Nova/Neutron telemetry wishlist mixing resource metrics with service-internal request, scheduler, RPC, notification, and utilization data, so it needs separate metric definitions before it can be implemented or closed.

Sharpz7:
Further update: PR #624 also covers the API-backed subset of #317 (hypervisor saturation, instance state/resource metrics, resource counts, and agent resource counts). #317 remains open only for service-internal or historical telemetry that needs separate metric definitions.

Sharpz7:
[...]

Issue #317, opened 2023-12-06
Looking for more Nova and Neutron metrics


NileshChandekar:
Any possibility to get this below list of metrics in the ``openstack-exporter`` list

### Nova (Compute Service):

~~~
nova_hypervisor_count:
nova_hypervisor_vcpus
nova_hypervisor_memory_used : Memory Usage on Hypervisors
nova_hypervisor_memory_used : Memory Usage on Hypervisors
nova_hypervisor_vcpus_used : CPU Usage on Hypervisors
nova_instance_errors: Number of instance errors.
nova_instance_deleted: Number of deleted instances.
nova_instance_build: Number of instances in the build state.
nova_instance_status: Status of instances (e.g., running, stopped).
nova_instance_vcpu: Number of virtual CPUs in use.
nova_compute_status: Indicates the status of Nova compute service
nova_compute_vcpus:
nova_compute_memory_bytes:
nova_compute_disk_available_bytes:
nova_instance_count:
nova_instance_active: Number of active instances ( running )
nova_instance_memory_mb: Memory usage of instances.
nova_instance_cpu_utilization: CPU utilization of instances.
nova_instance_memory_utilization: Memory utilization of instances.
nova_hypervisor_cpu_utilization: CPU utilization on the hypervisor.
nova_hypervisor_memory_utilization: Memory utilization on the hypervisor
nova_scheduler_running_filters: Number of filters applied by the scheduler.
nova_scheduler_filtered_hosts: Number of hosts filtered out during the scheduling process.
nova_scheduler_weighted_hosts: Number of hosts after applying weights in the scheduling process.
nova_conductor_rpc_request_duration_seconds: Duration of Nova conductor RPC requests.
nova_conductor_rpc_request_count: Count of Nova conductor RPC requests
nova_notifications: Monitor Nova notifications for events like instance creation, deletion, etc.
nova_api_error_count: Count of errors that occur in the Nova API.
nova_compute_error_count: Count of errors on the compute nodes.
nova_api_request_duration_Seconds :  Measures the duration of Nova API  requests
nova_api_request_count: Counts the Nova API requests
~~~
 
 
 
### Neutron (Networking Service):

~~~
neutron_network_active: Number of active networks.
neutron_subnet_active: Number of active subnets.
neutron_router_active: Number of active routers.
neutron_port_status: Status of ports (e.g., up, down).
neutron_floatingips: Number of floating IPs in use.
[...]

PR #624, opened 2026-09-01
Add metrics from issue #603

### LINKED ISSUE #83: add quota info for the resources
Is it possible to add quota metrics? we'd like to see how quota is used, and based on that we need the alert for quota usage.

### LINKED ISSUE #203: Snapshot counts by tenant
Is there a way to see snapshot count by tenant?  I need to monitor when snapshots are full by tenant.

### LINKED ISSUE #212: COE service list
Magnum (i.e. `container-infra` service) does offer the ability to list all of it's services and their current state over the API (similar to how Neutron exposes agents and Nova exposes services).

It would be useful to add this so that users can know if an agent is down.

### LINKED ISSUE #245: Object store metrics are only reported for the admin project
The Swift API supports listing containers only under specific accounts, so object store metrics only display containers of the project running the exporter (generally `admin`).

This affects `openstack_object_store_bytes` and `openstack_object_store_objects`.

### LINKED ISSUE #260: Export task_state of an instance in openstack_nova_server_status
An instance can transition through 42 additional task_states

https://github.com/openstack/nova/blob/master/nova/objects/fields.py#L988-L1030

As an `openstack-exporter` user I'd like to have `task_state` exposed in the `openstack_nova_server_status` metric so that I can monitor instances stuck in certain states.

### LINKED ISSUE #309: Missing metrics for OpenStack agent monitoring
The OpenStack exporter currently supports multiple agent_state metrics:

* `openstack_cinder_agent_state`
* `openstack_neutron_agent_state`
* `openstack_nova_agent_state`

We should add additional metrics for services such as Designate.

### LINKED ISSUE #314: extract `enabled` from `openstack_identity_project_info` as a new metric
`openstack_identity_project_info` contains `enabled` label that can take 2 different values `true` and `false` overtime, which makes duplicated metrics for periods where project changes status. This breaks queries like group_left which expects only single value for label matching. 

Example:
```
- expr: "last_over_time(openstack_cinder_volume_gb[2h])"
   record: "openstack_cinder_volume_expanded"

- expr: |
   label_replace(openstack_cinder_volume_expanded * 0 + 1, "image", "volume-$1", "id", "(.*)") *
       on (tenant_id) group_left(tenant_name)
          label_replace(
             label_replace(openstack_identity_project_info_expanded, "tenant_id", "$1", "id", "(.*)"
          ), "tenant_name", "$1", "name", "(.*)"
   )
   record: "openstack_cinder_volume"
```
If project state is changed in a 2h period, second rule will fail as it will have 2 results from same tenant_id in `openstack_identity_project_info_expanded`, one with `enabled=true` and other with `enabled=false`. If the metric is also used to count number of projects, it will be inaccurate. 

I could probably fix my expression somehow, but .. Would it make sense to extract it to a new metric like `openstack_project_enabled{id=}` which takes values 1/0 ?  

I know that it will be e breaking change, but I think that would make sense.

### LINKED ISSUE #347: Add ability to list users
Hi team!
Could we add a openstack_identity_users_info gauge as similar to what we've got with openstack_identity_projects_info?

We would like to use this information to join it with the nova_servers list and be able to generate a dashboard that shows the following structure for each node on Grafana (Using repeat option):

DUMMYNODE:
| ID | NAME | PROJECT | EMAIL |
|----|------|---------|-------|
| 78c8760d-0222-44c7-938b-17a6385e8436 | DUMMYVM | DUMMY_TENANT | dummy@tld.com |

### LINKED ISSUE #375: List Security Groups on instance or on ports
Hello,

Is it possible to have a list of security groups by instance or by ports

### LINKED ISSUE #473: Add nova/servergroups metrics
Hi,
It seems that the openstack exporter currently does not expose any metrics related to `servergroups`.

Could you please consider adding support for it?

### LINKED ISSUE #479: Add Manila share quota usage and limit metrics
Currently, this project already exposes several metrics to track quota usage across different OpenStack components, such as Cinder (`openstack_cinder_limits_volume_used_gb`) and Nova (`openstack_nova_limits_memory_used`).

It would be equally valuable to have similar metrics for Manila shares, for example:

- `openstack_sharev2_limits_shares_max_gb`
- `openstack_sharev2_limits_shares_used_gb`
- `openstack_sharev2_limits_shares_max_instances`
- `openstack_sharev2_limits_shares_used_instances`

These metrics could then be integrated into monitoring systems to trigger alerts whenever Manila share usage approaches the configured quota limits.

### LINKED ISSUE #531: If possible add a label to openstack_cinder_volume_status for os-vol-host-attr:host if returned
Hi,

By adding this, it would make it possible to filter/group volume per host@backend#pool, thus have a better understanding on how the cinder backend is utilised. It would be possible to report the overcommit ratio and distribution between the backend pools.

The block-storage API for get volume does return the property if available.
https://docs.openstack.org/api-ref/block-storage/v3/#volumes-volumes

```
$ openstack volume show temp9                                                                                                                                 
+--------------------------------+---------------------------------------------------------------------+
| Field                          | Value                                                               |
+--------------------------------+---------------------------------------------------------------------+                                                                                 | attachments                    | []                                                                  |
| availability_zone              | nova                                                                |
| bootable                       | false                                                               |
| consistencygroup_id            | None                                                                |
| created_at                     | 2026-02-02T20:01:10.000000                                          |
| description                    | None                                                                |
| encrypted                      | False                                                               |
| id                             | 5fc7d35a-cdc2-444e-bbff-************                                |
| migration_status               | None                                                                |
| multiattach                    | False                                                               |
| name                           | temp9                                                               |
| os-vol-host-attr:host          | juju-******-0-lxd-3@cinder-netapp-********#********_openstack_vol03 |
| os-vol-mig-status-attr:migstat | None                                                                |
| os-vol-mig-status-attr:name_id | None                                                                |
| os-vol-tenant-attr:tenant_id   | 8bdf1ff346844448a7c9dd3d20672a1b                                    |
| properties                     |                                                                     |
| replication_status             | None                                                                |
| size                           | 1000                                                                |
| snapshot_id                    | None                                                                |
| source_volid                   | None                                                                |
| status                         | available                                                           |
| type                           | ********-openstack                                                  |
| updated_at                     | 2026-02-02T20:01:10.000000                                          |
| user_id                        | 25dceae911f9449d91**************                                    |
+--------------------------------+---------------------------------------------------------------------+
```

Regards,

Sharpz7:
Part of #603 and #317.

Adds tenant snapshot counts, Cinder volume host and quota labels, Nova task states, server groups, hypervisor saturation and aggregate compute capacity metrics, instance resource/state metrics, Keystone project/user metrics, Nova/Neutron security-group and resource counts, Magnum and Designate service health, Manila share quota metrics, and Swift project-scoped container metrics. Adds unit coverage and integration assertions.

Closes #83
Closes #203
Closes #212
Closes #245
Closes #260
Closes #309
Closes #314
Closes #347
Closes #375
Closes #473
Closes #479
Closes #531


Sharpz7:
One small follow-up: `Deleted` is not currently exposed in gophercloud’s `servers.ListOpts`, so this PR keeps a local query type for `deleted=true`. This could be moved into gophercloud in a future change, then simplified here.

